PR #4, opened 2020-05-01
Install script

stepanstipl:


NargiT:
must be run with sudo on ubuntu

stepanstipl:
Ah, thanks for catching this. I'm just thinking about what would be preferable way then:
- Just use sudo in readme?
- Install in different path... where? (seems like `~/bin` is recommended for user binaries?, but then it's not in `$PATH`)
- The script should try to use sudo?

stepanstipl:
The install script will now try to use sudo if destination dir is not writeable, and if it won't help, it will end with a hopefully helpful error message:
```
ERROR: Target directory (/usr/local/bin) is not writable (destination can be changed using $TARGET_DIR variable)
```

PR #9, opened 2020-07-13
Added first test

david-doit-intl:
Will add Github Actions testing, this is just making sure you are ok with the flow.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -19,7 +19,7 @@ jobs:
     - name: Test
       run: |
         scripts/alpine-setup.sh
-        GOOS=linux GOARCH=amd64 make test
+        make test
   build:
     name: Build
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yaml
```diff
@@ -7,6 +7,22 @@ on:
...
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - name: Test
+      run: |
+        apk add git
+        go get github.com/rakyll/statik
+        go generate
+        export CGO_ENABLED=0
+        go test ./...
```
stepanstipl, line 25: Can we use the same `scripts/alpine-setup.sh` script (I know it sets up upx too which is not needed), as in the Build step, to install bits in the image and then add this as `make test` target to Makefile, so that the tests can easily be run locally as well? (this target can depend on generate, and Makefile already has `CGO_ENABLED=0`)

david-doit-intl: Sure :)

david-doit-intl:
@stepanstipl all done, please review.

stepanstipl:
Thanks, merged & I've made the `test` check required for future PRs 🌈  👍

PR #23, opened 2020-08-12
Feature/docker

david-doit-intl:
Added Docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM golang:1.13.9-alpine3.11 as builder
+
+WORKDIR /src
+
+COPY go.mod .
+COPY go.sum .
+COPY scripts scripts
+
+RUN scripts/alpine-setup.sh 
+
+RUN go mod download
+
+COPY . .
+
+RUN make all
+
+FROM scratch
+
+COPY --from=builder /src/bin/kubent-linux-amd64 /app/kubent
+
+WORKDIR /app
+
+CMD ["./kubent"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+#
+bin/
+release-artifacts/
+generated/
+debug.test
+
+# Dependency directories (remove the comment below to include it)
+# vendor/
+
+# Secrets
+.envrc
+
+# Editor files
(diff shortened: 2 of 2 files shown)


david-doit-intl:
@stepanstipl should we release this as well as the go binary?

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY go.sum .
+COPY scripts scripts
+
+RUN scripts/alpine-setup.sh 
+
+RUN go mod download
+
+COPY . .
+
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 make all
```
stepanstipl, line 15: just `make all` is sufficient to avoid duplication, the other bits are in the Makefile/auto-detected

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM golang:1.13.9-alpine3.11 as builder
+
+WORKDIR /go/github.com/doitintl/kube-no-trouble
```
stepanstipl, line 3: can we move this to smth like just `/src`? This is not a "correct" path with old `$GOPATH` scheme anyway (missing `src/` bit), and go historically has split-brain behavior with $GOPATH and modules, I would prefer to make it clear

Dockerfile
david-doit-intl: sure will change

Dockerfile
david-doit-intl: sure :)

david-doit-intl:
I don't think tini is needed. Would like to keep this part simiple. 

- We can use google's gcloud container as the final image, so we don't have the headache: google/cloud-sdk instead of scratch, if we want to make it compatable with GKE
- kubeconfig can be added via volume, if needed, or readout via the commandline, we don't have to support everything at once in the docker image
- Yeah, again we can take small steps, we can decide what to support and when, nice thing about having the binary in scratch is that it can be used in anyone of these places.

stepanstipl:
Yep, thanks for the changes, let's open & this in separate issues if needed. I'll create one for building & pushing the image somewhere, as I would like to exercise the code if it's in the repo ;).

PR #24, opened 2020-08-12
Feature/clean up repo

david-doit-intl:


Makefile
```diff
@@ -51,54 +52,63 @@ _ := $(foreach exec,$(REQ_BINS), \
...
 
-all:        build pack release-artifacts                            ## Clean, build and pack
+## Clean, build and pack
+all: build pack release-artifacts
 .PHONY: all
 
-help:                                                               ## Prints list of tasks
-	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z0-9_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' Makefile
+## Prints list of tasks
+help:
```
stepanstipl, line 60: make target commans have to start with tabs, otherwise:
```
$ gmake help                                                                                                                                                                   0
gmake: Nothing to be done for 'help'.
```

scripts/alpine-setup.sh
```diff
@@ -12,6 +12,7 @@ apk add --update --no-cache \
 	make \
 	tar
 
+mkdir -p bin generate
```
stepanstipl, line 15: `generate` -> `generated`, but let's lave this func in the Makefile, as it will work even if someone doesn't use the install script.

Makefile
```diff
@@ -51,54 +52,63 @@ _ := $(foreach exec,$(REQ_BINS), \
...
-pack: $(PACKED_BINS)                                                ## Pack binaries with upx
+## Pack binaries with upx
+pack: $(PACKED_BINS)
 .PHONY: pack
 
 $(PACKED_DIR)/%-$(BIN_ARCH): $(BIN_DIR)/%-$(BIN_ARCH)
 	mkdir -p $(PACKED_DIR)
 	$(UPXCMD) --brute -f -o "$@" "$<" \
-	  && touch "$@"
+		&& touch "$@"
```
stepanstipl, line 88: This is strange formatting

Makefile
```diff
@@ -51,54 +52,63 @@ _ := $(foreach exec,$(REQ_BINS), \
...
 	test -z "$$(go fmt ./...)"
 .PHONY: test-fmt
 
-clean:                                                              ## Clean build artifacts
-	rm -rf generated/*	
-	rm -rf bin/*
+## Clean build artifacts
+clean:
+	rm -rf generated
+	rm -rf bin
```
stepanstipl, line 112: I think removing the dirs instead of files, is ok, but let's use the variables `$(GENERATE_DIR) $(BIN_DIR)` since we have them

Makefile
```diff
@@ -51,54 +52,63 @@ _ := $(foreach exec,$(REQ_BINS), \
...
+	mkdir -p $(GENERATE_DIR) $(BIN_DIR)
```
stepanstipl, line 113: Related to the other changes around this -  👍  for deleting dirs, seems cleaner. But instead of making the new ones and several places (here & install script), what about moving this logic to the steps that actually need them.

So `mkdir -p $(GENERATE_DIR)` to `generated/*:` target and keeping the `mkdir -p $(BIN_DIR)` where it is.

Makefile
```diff
@@ -51,54 +52,63 @@ _ := $(foreach exec,$(REQ_BINS), \
...
+  @awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z0-9_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' Makefile
 .PHONY: help
 
-build: $(BINS)                                                      ## Build binary
+## Build binary
+build: $(BINS)
 .PHONY: build
 
 $(BIN_DIR)/%-$(BIN_ARCH): generated/* $(SRC)
-	mkdir -p $(BIN_DIR)
```
stepanstipl, line 65: I would keep this here to avoid duplication in Makefile & install script.

stepanstipl:
OK this awk would work with new lines (only slighly more cryptic 😉 ):
```shell
@awk 'BEGIN {FS=":"} /^## .*/,/^[a-zA-Z0-9_-]+:/ { if ($0 ~ /^## /) { desc=substr($0, 4) } else { printf "\033[36m%-30s\033[0m %s\n", $1, desc } }'
```

Makefile
david-doit-intl: Fixed!

Makefile
david-doit-intl: Yes! Done :)

Makefile
david-doit-intl: Let me know if it works now ;)

stepanstipl: Nope :( there still seems to be a space (instead of tab) in front of the `  @awk ...` line

david-doit-intl: honestly this feels alittle insane, makefile feels super sensitive.

stepanstipl: I think it's just Make syntax (always been :)), seems like it's configurable in GNU make with `.RECIPEPREFIX` [1], still will take only 1 char 😄 . I agree it's a bit odd, but this would bring to a long discussion "Yeah Make is not that great"... but what is better and present everywhere by default? :)

[1]: https://www.gnu.org/software/make/manual/html_node/Rule-Syntax.html

david-doit-intl: I think its best if we just print out the help details, rather than parsing our own makefile.

Issue #25, opened 2020-08-13
panic: runtime error: invalid memory address or nil pointer dereference


san360:
Error on WSL1 windows OS

`❯ kubent
2:44PM INF >>> Kube No Trouble kubent <<<
2:44PM INF version 0.3.0 (git sha dev)
2:44PM INF Initializing collectors and retrieving data
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1154c15]

goroutine 1 [running]:
github.com/doitintl/kube-no-trouble/pkg/collector.(*ClusterCollector).Name(0x0, 0x1823920, 0xc000361b80)
        <autogenerated>:1 +0x5
main.storeCollector(0x182ac00, 0x0, 0x1823920, 0xc000361b80, 0x2331290, 0x0, 0x0, 0x0, 0x0, 0x16649b5)
        /__w/kube-no-trouble/kube-no-trouble/cmd/kubent/main.go:40 +0x151
main.initCollectors(0xc000223630, 0x16649b5, 0x2b, 0xc0002ce540)
        /__w/kube-no-trouble/kube-no-trouble/cmd/kubent/main.go:51 +0x3a7
main.main()
        /__w/kube-no-trouble/kube-no-trouble/cmd/kubent/main.go:88 +0x31c`

stepanstipl:
Hi @san360 thanks for reporting this. I'll certainly try to look into this, but tbh. it might take some time, I'm not familiar with WSL at all ;). Are you able to try with WSL2 - partially out of curiosity on my side, no idea if that might help, but my colleague mentioned there were many improvements/bugfixes since the WSL1.

As a workaround, I would suggest you try to use docker image. We're not publishing it anywhere yet (tbd soon), but just recently added it to the repo, so just building the image from Dockerfile in the root of this repo should work.

manute:
I have the same error, in mac osx and **version 0.3.0** and I think is due to a bad kube config, so I've created a [PR](https://github.com/doitintl/kube-no-trouble/pull/32) to avoid the panic and the error is showing up like: 
```
> ./bin/kubent                                                                                                                                                                                (use1-test-4/default)
3:32PM INF >>> Kube No Trouble `kubent` <<<
3:32PM INF version dev (git sha dev)
3:32PM INF Initializing collectors and retrieving data
[...]

PR #32, opened 2020-09-01
Fixing panic when collector name is nil

manute:
https://github.com/doitintl/kube-no-trouble/issues/25


stepanstipl:
Thanks @manute, this seems to indeed fix the error mentioned in #25.

@david-doit-intl I think we should cover this case with a test, I'll create an issue

Issue #35, opened 2020-09-03
Add test to cover collector logic


stepanstipl:
Cover #30 - kube fake client should help here - https://gianarb.it/blog/unit-testing-kubernetes-client-in-go

Issue #33, opened 2020-09-03
Add test for storeCollector - err in particular


stepanstipl:
Add tests for storeCollector, in particular cover when collector fails to be initialised  #25, #32

Issue #30, opened 2020-08-26
kubent 0.3.0 NOT reporting 1.16 usage of deprecated apis that ARE reported by kubent 0.2.0


gfahyhpe:
Version 0.2.0 reports ...

```
10:33AM INF >>> Kube No Trouble `kubent` <<<
10:33AM INF version 0.2.0 (git sha b9b45f347c7e67cf488d8dd6e20c8b848c595a3e)
...
10:33AM INF Loaded ruleset name=deprecated-1-16.rego
10:33AM INF Loaded ruleset name=deprecated-1-20.rego
..
>>> 1.16 Deprecated APIs <<<
------------------------------------------------------------------------------------------
KIND         NAMESPACE     NAME                          API_VERSION
DaemonSet    <REDACTED>   <REDACTED>        extensions/v1beta1
Deployment   <REDACTED>   <REDACTED>         extensions/v1beta1
Deployment   <REDACTED>   <REDACTED>         apps/v1beta2
Deployment   <REDACTED>   <REDACTED>         extensions/v1beta1
Deployment   <REDACTED>   <REDACTED>         extensions/v1beta1
Deployment   <REDACTED>   <REDACTED>         extensions/v1beta1
Deployment   <REDACTED>   <REDACTED>         extensions/v1beta1
__________________________________________________________________________________________
...
>>> 1.20 Deprecated APIs <<<
------------------------------------------------------------------------------------------
A long list similar to that for 1.16 above
```

BUT
Latest version 0.3.0 for same cluster reports

```
$ kubent_30 -k kubeconfig-readonly.yaml
10:40AM INF >>> Kube No Trouble `kubent` <<<
10:40AM INF version 0.3.0 (git sha dev)
...
10:40AM INF Loaded ruleset name=deprecated-1-16.rego
10:40AM INF Loaded ruleset name=deprecated-1-22.rego
________________________________________________________________________________________
>>> Deprecated APIs removed in 1.22 <<<
------------------------------------------------------------------------------------------
KIND      NAMESPACE   NAME                          API_VERSION          
Ingress   <REDACTED> <REDACTED>             extensions/v1beta1  
```

showing NOTHING for 1.16 Deprecated APIs

Am I doing something wrong ?

jmcdon01:
Yes I have noted this also. Something has changed in what it's collected/reported.

stepanstipl:
[...]

PR #37, opened 2020-09-09
More tests

stepanstipl:
More tests, namely to cover `storeCollector` and introducing use of k8s fake client in cluster collector.

Related to #33, #25, #32, #35, #30.


david-doit-intl (approved):

david-doit-intl:
👍

Issue #31, opened 2020-08-28
CI mode


estahn:
The tool should provide a way for CI to fail. This is typically done by returning a non-zero exit code.

stepanstipl:
Thanks @estahn, this is good suggestion. I was thinking about this earlier, and initially decided to reserve non-zero return codes to actual failure - ie. if there's an error.

It should be fairly simple to do basic test in CI, as "log" outputs go to stderr, and only findings go to stdout. Following should give you the non-zero if there are any findings:
```
test -z "$(kubent)"
```
Would that cover your use case?

There's also json output mode (`-o json`) that should allow for amore convenient further processing of results.


Alternatively, what behaviour would you expect in terms of differentiating between error and deprecated resources found?

Djiit:
FWIW, Trivy has this `exit-code` parameter (https://github.com/aquasecurity/trivy#specify-exit-code). The default is 0, but you can tell Trivy to return 1 (or whatever) if it make sense. If a real error is encountered, this parameter is ignored.

stepanstipl:
This was closed via #49, `-e` flag will now make kubent to return non-zero (`200`) exit code in case issues are found. Also [Use in CI](https://github.com/doitintl/kube-no-trouble#use-in-ci) section was added to the docs mentioning this new flag and some other examples. Finally, we've added tests to make sure the error-code behaviour is as expected and will stay so. 🚀 . Thanks @estahn and @Djiit for you input on the topic! 👍 

_(This is now only in master, but I expect to cut a new release probably early next week)._

PR #46, opened 2020-10-21
Empty json output to array

stepanstipl:
This PR makes an empty output (i.e. no issues found) in json format to be an empty array  (instead of null). This is more friendly for CI use (#31), such as with jq.

Before:
```
$./bin/kubent-darwin-amd64 -o json                                                                                                                       2
2:36PM INF >>> Kube No Trouble `kubent` <<<
...
...
null
```

After:
```
$./bin/kubent-darwin-amd64 -o json                                                                                                                     130
3:09PM INF >>> Kube No Trouble `kubent` <<<
...
...
[]
```

### DIFF
--- a/pkg/judge/rego.go
+++ b/pkg/judge/rego.go
@@ -66,7 +66,7 @@ func (j *RegoJudge) Eval(input []map[string]interface{}) ([]Result, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


david-doit-intl (approved):
LGTM

PR #49, opened 2020-10-22
Add --exit-error (-e) flag

### LINKED ISSUE #31: CI mode

stepanstipl:
This PR adds `--exit-error` (`-e`) flag for convenience. When enabled this will make `kubent` return non-zero exit code in case any issues are found. Also, tests for exit codes are added. 

Fixes #31.

### DIFF
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -22,6 +22,12 @@ var (
 	gitSha  string = "dev"
 )
 
+const (
+	EXIT_CODE_SUCCESS      = 0
+	EXIT_CODE_FAIL_GENERIC = 1
+	EXIT_CODE_FOUND_ISSUES = 200
+)
+
 func getCollectors(collectors []collector.Collector) []map[string]interface{} {
 	var inputs []map[string]interface{}
 	for _, c := range collectors {
@@ -70,6 +76,7 @@ func initCollectors(config *config.Config) []collector.Collector {
 }
 
 func main() {
+	exitCode := EXIT_CODE_FAIL_GENERIC
 
 	config, err := config.NewFromFlags()
 	if err != nil {
@@ -114,4 +121,11 @@ func main() {
 	if err != nil {
 		log.Fatal().Err(err).Msg("Failed to print results")
 	}
+
+	if config.ExitError && len(results) > 0 {
+		exitCode = EXIT_CODE_FOUND_ISSUES
+	} else {
+		exitCode = EXIT_CODE_SUCCESS
+	}
+	os.Exit(exitCode)
 }
--- a/cmd/kubent/main_test.go
+++ b/cmd/kubent/main_test.go
@@ -2,6 +2,9 @@ package main
 
 import (
 	"errors"
+	"os"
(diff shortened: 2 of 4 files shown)


david-doit-intl:
Shouldn't this exit with 0, by default?

stepanstipl:
> Shouldn't this exit with 0, by default?

It does, this is kinda for convenience so that you don't have to check this explicitly (#31), it's described better in committed docs, sorry if it wasn't clear from the PR:

> kubent will by default return 0 exit code if the program succeeds, even if it finds deprecated resources, and non-zero exit code if there is an error during runtime. Because all info output goes to stderr, it's easy to check in shell if any issues were found:

... 

> You can also use --exit-error (-e) flag, which will make kubent to exit with non-zero return code (200) in case any issues are found.

cmd/kubent/main.go
```diff
@@ -70,6 +75,7 @@ func initCollectors(config *config.Config) []collector.Collector {
 }
 
 func main() {
+	exitCode := EXIT_CODE_SUCCESS
```
david-doit-intl, line 78: I think it should be 200, by default incase of silient errors :)

cmd/kubent/main_test.go
```diff
@@ -85,3 +88,53 @@ func TestStoreCollectorError(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			var ee *exec.ExitError
+
+			cmd := exec.Command(os.Args[0], "-test.run=TestMainExitCodes")
+			cmd.Env = append(os.Environ(), "TEST_EXIT_CODE=1", "TEST_CASE="+strconv.Itoa(i))
+			err := cmd.Run()
+
+			if tc.expected == 0 && err != nil {
+				t.Errorf("expected to succeed with exit code %d, failed with %v", tc.expected, err)
+			} else if tc.expected != 0 {
```
david-doit-intl, line 127: Break up the nested logic here.

cmd/kubent/main.go
```diff
@@ -22,6 +22,12 @@ var (
 	gitSha  string = "dev"
 )
 
+const (
+	EXIT_CODE_SUCCESS      = 0
+	EXIT_CODE_FAIL_GENERIC = 1
```
david-doit-intl, line 27: Good!

cmd/kubent/main.go
stepanstipl: So we set exit code to be 1 (generic error) by default & only change it to 0 or 200 based on the flag if we successfully get to the end.

david-doit-intl (approved):
Better tests structure now too!

cmd/kubent/main_test.go
stepanstipl: @david-doit-intl  Pls. have a look, I *think* it's now a bit more readable by using `t.Failf` and un-nesting the logic.

david-doit-intl:
We should probably change all the tests to use the t.Fatalf, but thats for annother pr

PR #53, opened 2020-10-23
Pre-commit

david-doit-intl:

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -7,6 +7,12 @@ on:
     - "*"
   pull_request:
 jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - uses: pre-commit/action@v2.0.0
   test:
     name: test
     runs-on: ubuntu-latest
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,3 @@ debug.test
 # Editor files
 .vscode
 .swp
-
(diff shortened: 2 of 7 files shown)


stepanstipl (approved):
Thanks @david-doit-intl .

1) The doc improvements are nice 👍 

2) The test coverage 👍 - it would be nice to expose this somehow more visibly and check that we're not getting worse with any single commit :)... but this is a good start.

3) I have some concerns about the pre-commit func. In particular:

- a) It expects/installs stuff on host system which I find a bit dangerous, and cluttering - ie. pre-commit is python, but will install go (such as checkmake). I didn't dig into this too much, but as someone who's trying to control what and in which way is installed on my system, this makes me somewhat uncomfortable. I expect this can "break" in many ways, depending on your OS and other bits.
- b) Reproducibility of the builds - because of environment dependencies, and referencing `master` versions of those hook repos., we might get different results.
- c) Quality and usefulness of the individual hooks - like for example the ` checkmake` one doesn't seem to do much.

Overall it seems like a potentially complex system with a lot of dependencies.

Also re. the build reproducibility, I would like to keep easy parity of being able to execute all tests via `make`, both locally and in ci. There's a pattern, which I like (and would assume we have proper build image, currently `golang` one). This allows you to easily run any make step both locally and/or from inside the container), as well from outside by prefixing it with `docker-`.
```
docker-%:
	docker run -it --rm -e TF_VAR_project $(DOCKER_IMAGE) $*
```


Finally, I really like the idea of pre-commit (just not the implementation so much), and realised how tired am I of writing same CI bits over and over again :). In longer-term, I started looking into nix and it's shell functionality (check 
https://github.com/hashicorp/waypoint/blob/main/shell.nix for an example how this can be used) as a better option to purely docker based reproducible dev/build environments. And considering building a shared library of "CI" scripts... but this might take a while.

So on a pragmatic note, the results are good -> let's merge this as it is 👍  and improve later :).

PR #55, opened 2020-10-28
Dep autorest

stepanstipl:
Updated autorest dependency to resolve the ambiguous import (Azure/go-autorest#528) issue and get `go mod tidy` working. Also committed generated files (we should add some check these are up-to-date) in order to make `go mod tidy` and Dependabot combo happy (fixes https://github.com/doitintl/kube-no-trouble/network/updates/68342830).


david-doit-intl (approved):

david-doit-intl:
I just not sure about the generated files thats all

stepanstipl:
> I just not sure about the generated files thats all

As in committing these to git repo?

stepanstipl:
@david-doit-intl  Just thinking about this, there seem to be 3 options:

a) Not using Dependabot at all -> maybe later move to a custom solution, but for now, I would prefer to try out of the box one.

b) Use, but disable `tidy` option - I'm a bit afraid this would lead to accumulation of dependencies, but maybe that can be implemented as a separate action/step

c) Use, and commit generated files (this PR)


I'm kinda open to trying the option b) first, if you find it preferable.

stepanstipl:
_just for record - concern is about generated files in generate, not whether to commit or not in git -> let's discuss separately_

PR #56, opened 2020-10-28
build: Add check for commit message format

stepanstipl:
This PR adds check and docs for commit message format.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -21,7 +21,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v1
       with:
-        fetch-depth: 1
+        fetch-depth: 0
     - name: Test
       run: |
         scripts/alpine-setup.sh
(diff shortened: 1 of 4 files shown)


david-doit-intl (approved):

david-doit-intl:
👍

david-doit-intl:
we could do this inside a python docker image with commitizen installed instead of bash script, but its a such a minor point. I think this is fine.

stepanstipl:
> we could do this inside a python docker image with commitizen installed instead of bash script, but its a such a minor point. I think this is fine.

Yep, I was looking at a couple of tools ([commitizen](https://github.com/commitizen/cz-cli), [commitlint](https://github.com/conventional-changelog/commitlint), [conform](https://github.com/talos-systems/conform), [go-gitlint](https://github.com/llorllale/go-gitlint)), and most of them were missing some config options or it ends-up being regexp anyway, + not a fan of pulling nodejs & all the dependencies. So, in the end, this seemed like the easiest. But agree overall it would be nice to have some good pattern to move this externally... some ideas in #53.

PR #70, opened 2020-11-11
ci: Create Nightly releases

stepanstipl:
This should create a nightly build if there are any changes... currently figuring out a way to test.


david-doit-intl (approved):

stepanstipl:
Lots of trial/error with GH actions, but it seems to be working now. Let's put it to a real test :)

PR #72, opened 2020-11-27
Removing whitespace at the end of the ruleset 1.16 msg

manute:
Hi again 👋 ! 

I've found that the msg of the 1.16 deprecation ruleset has an extra whitespace at the end, I think it's a little bit annoying for consuming the json output with `jq`  so I removed it in this PR. 

Let me know if that's ok.

Thanks!

### DIFF
--- a/rules/deprecated-1-16.rego
+++ b/rules/deprecated-1-16.rego
@@ -9,7 +9,7 @@ main[return] {
 		"Namespace": get_default(resource.metadata, "namespace", "<undefined>"),
 		"Kind": resource.kind,
 		"ApiVersion": old_api,
-		"RuleSet": "Deprecated APIs removed in 1.16 ",
+		"RuleSet": "Deprecated APIs removed in 1.16",
 	}
 }
 


david-doit-intl:
Hi Manute,

Thanks for the contribution. I think there is some rule which requires there to be white space there. @stepanstipl wdyt?

stepanstipl:
I think this should be completely fine. Thanks @manute 👍 
Just needs a rebase and squash/fix the commit message to match the [required style](https://github.com/doitintl/kube-no-trouble#commit-messages).

manute:
@stepanstipl  I addressed your feedback, thanks for the review.

stepanstipl (approved):
Great, thanks @manute for your contribution! 👍  Looks good, I'm going to merge this.

Before we cut a release, there are now nightly builds in place, so this change should be included in the next nightly build - these are listed as pre-releases on the GH  [releases] (https://github.com/doitintl/kube-no-trouble/releases) page.

Thanks again! 🎉

PR #79, opened 2020-12-11
build: Allow hyphens in tags - for nightly

stepanstipl:
🤞 this will finally allow the nightly builds to work


david-doit-intl (approved):
Perfect

Issue #67, opened 2020-11-04
helm deployments return undefined namespaces


jbosm1:
for some reason helm deployments seem to return namespaces as undefined

stepanstipl:
Hi @jbosm1,

thanks for reporting the issue. I'll look into that.

Would you be able to paste here the output from kubent? Thanks.

urmet:
I see the same thing. the deployments are created via helm2 that is controlled by gitlab kubernetes integration
```urmet@superkiisu ~ $ kubent
6:33PM INF >>> Kube No Trouble `kubent` <<<
6:33PM INF version 0.3.2 (git sha 919129b596890475965cda2b972cb6fded71f40b)
6:33PM INF Initializing collectors and retrieving data
6:33PM INF Retrieved 28 resources from collector name=Cluster
6:33PM INF Retrieved 67 resources from collector name="Helm v2"
6:33PM INF Retrieved 99 resources from collector name="Helm v3"
6:33PM INF Loaded ruleset name=deprecated-1-16.rego
6:33PM INF Loaded ruleset name=deprecated-1-22.rego
__________________________________________________________________________________________
>>> Deprecated APIs removed in 1.16  <<<
------------------------------------------------------------------------------------------
KIND         NAMESPACE     NAME                                    API_VERSION          
Deployment   <undefined>   ingress-nginx-ingress-controller        extensions/v1beta1   
Deployment   <undefined>   ingress-nginx-ingress-default-backend   extensions/v1beta1   ```

UberKuber:
I think this occurs if the namespace wasn't specified in the metadata of the deployment.

meringu:
Is it safe to default to the tiller namespace here?

meringu:
Looks like we have access to the release namespace here. I'll try write up a fix tomorrow

meringu:
^ The linked PR solves this problem in my testing.

stepanstipl:
This should be fixed now, big thanks to @meringu  🎉 👏 for the PR, so should be part of next nightly release. 🎉

meringu:
Hey @stepanstipl. Thanks for the merge. Any chance we could cut a 0.3.3 release?

poornimaram88:
Hi Will there be a 0.3.3 release soon? I want to use this fix as many of my helm 2 ns are listed as undefined.

stepanstipl:
[...]

PR #95, opened 2021-01-28
fix: Default the Helm manifest namespace to the release namespace

meringu:
closes: #67

### DIFF
--- a/pkg/collector/helm2.go
+++ b/pkg/collector/helm2.go
@@ -2,6 +2,7 @@ package collector
 
 import (
 	"fmt"
+
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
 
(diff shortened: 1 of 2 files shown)


david-doit-intl:
LGTM, lets let @stepanstipl have a look.

stepanstipl:
Thanks @meringu, I think this great! 👍  I'm reasonably confident this will work for Helm v3, as the release info is stored in the same namespace where the release is deployed.

However for v2, won't this always be the namespace where the Tiller is deployed (I think `kube-system` by default)?

TL;DR: You mentioned it worked in your setup in #67, were you able to test this with Helm v2?

meringu:
Yes we have a very large helmv2 deployment still.

Tiller can go in any namespace, it only defaults to kube-system. This is configured with the --tiller-namespace flag. However, when you install a release you can override the namespace with the --namespace flag.

This seems to report correctly over our hundreds of helmv2 deprecation warnings. The release namespace ends up being the default namespace to deploy the resources, just like helmv3.

stepanstipl (approved):
OK, thanks for confirming this, sounds good. 👍

Issue #87, opened 2021-01-09
VolumeSnaphosts are moving to v1


stepanstipl:
https://kubernetes.io/blog/2020/12/10/kubernetes-1.20-volume-snapshot-moves-to-ga/

Research and consider adding a warning for old API versions (alphas v1/v2 + beta)

dja-fr:
Hello,

Maybe a little refactor to show in addition of removed API the list of currently deprecated API but without removal announce ?

On rules folder something like that ? :

```
rules/
  1-16-deprecated.rego
  1-16-removed.rego
  1-20-deprecated.rego
  1-21-deprecated.rego
  1-22-deprecated.rego
  1-22-removed.rego
 
```
Two file per K8s release seems to be easier to maintain rather than 1 global file with deprecation list.

and for console output 

```
__________________________________________________________________________________________
>>> Deprecated APIs annoucement in 1.20 <<<
------------------------------------------------------------------------------------------
KIND          NAMESPACE     NAME                          API_VERSION                     REPLACE_WITH (SINCE)
VolumeSnapshot <undefined>   test-volume   snapshot.storage.k8s.io/v1beta1   snapshot.storage.k8s.io/v1 (1.20+)
__________________________________________________________________________________________
>>> Deprecated APIs removed in 1.22 <<<
------------------------------------------------------------------------------------------
KIND          NAMESPACE     NAME                          API_VERSION                     REPLACE_WITH (SINCE)
TokenReview   <undefined>   test-deprecated-TokenReview   authentication.k8s.io/v1beta1   authentication.k8s.io/v1 (1.19+)
```

stepanstipl:
Thanks, @dja-fr for the idea, I think it would make sense to have a separate group for announced deprecated APIs, when it's not clear in which version these will be removed.

I'm thinking about two options:

a) As you described, have additional set for each K8s version, then how would the progression over time look like:

- i) API appears in the announced group (like VolumeSnapshot in 1.20)
- ii) Once it's clear when removed, API is moved to corresponding deprecated (like VolumeSnapshot to 1.25 - just an example), I guess we don't want to have rules twice ;)
- iii) the "1.20 - announced" group will eventually disappear

b) Have a generic group for deprecated APIs, without pinning to a specific version

- i) API appears in the generic "deprecated" group
- ii) Once clear when removed, it will move to corresponding deprecated


[...]

PR #97, opened 2021-02-03
feat: Add deprecation rules without removal announcement

dja-fr:
Generic deprecated.rego file added. It contains api depreciation annoucement list without target version for removal.

Aim to fix https://github.com/doitintl/kube-no-trouble/issues/87

### DIFF
--- a/rules/deprecated.rego
+++ b/rules/deprecated.rego
@@ -1,0 +1,114 @@
+package deprecated
+
+main[return] {
+	resource := input[_]
+	api := deprecated_resource(resource)
+	return := {
+		"Name": resource.metadata.name,
+		# Namespace does not have to be defined in case of local manifests
+		"Namespace": get_default(resource.metadata, "namespace", "<undefined>"),
+		"Kind": resource.kind,
+		"ApiVersion": api.old,
+		"ReplaceWith": api.new,
+		"RuleSet": "Deprecated APIs without removal annoucement",
+		"Since": api.since,
+	}
+}
+
+deprecated_resource(r) = api {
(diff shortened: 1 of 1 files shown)


stepanstipl:
Thanks @dja-fr! This is great 🎉. I'm only wondering if all of these should be enabled by default. I imagine in a lot of cases people are using this tool with older versions, as 1.19/1.20 are still relatively new, and I would not like to overwhelm users.

Currently, my thinking is to add some mechanism that would display only relevant deprecations by default. I.e. if we can detect K8s minor version (`m`), we would only show depreciations that have replacements available in `m+1`.

What do you think about this?

stepanstipl:
Hi @dja-fr I haven't heard back about this. I believe most are (or should be, as per my comments) now covered in #144 - by the 1.22 & 1.25 rules. Please check there if it covers everything, and/or if you have any additional comments.

However some of these are not though, I believe the following:
- `BootstrapTokenString`
- `HorizontalPodAutoscaler`

Do you have some source of info about these?

Also, I'm not sure about all the `*List*` resources, but I believe these are never directly used in the cluster, but rather helper generated resources. Do you think we need to include those?

stepanstipl:
These have now been covered by separate PRs as listed in #135 (also in the meantime, these have been scheduled for removal in 1.22 & 1.25). Please check the latest nightly, or the next `0.4.1` (to be out soon).

I haven't head back on the 2 resources:

> However some of these are not though, I believe the following:
> - `BootstrapTokenString`
> - `HorizontalPodAutoscaler`
> 
> Do you have some source of info about these?

But please open separate individual issues with a link to some reference and we can handle those separately.

Thanks for your help 👍

Issue #94, opened 2021-01-27
networking.k8s.io/v1beta1 should be flagged deprecated also


jmcdon01:
Since 1.19 has moved ingress to v1 could we alert on both v1beta1 strings? 

		"Ingress": {
			"old": ["extensions/v1beta1"],
			"new": "networking.k8s.io/v1beta1"

https://v1-19.docs.kubernetes.io/docs/setup/release/notes/#deprecations-and-removals
_Ingress and IngressClass resources have graduated to networking.k8s.io/v1. Ingress and IngressClass types in the extensions/v1beta1 and networking.k8s.io/v1beta1 API versions are deprecated and will no longer be served in 1.22+_

Once again thanks for the fantastic tool!

mausch:
Hi, this was closed 2021-06-11 but the latest release 0.5.1 (2021-10-26) doesn't seem to have it?

stepanstipl:
Hi @mausch,

I've checked with `0.5.1` in particular, and it seems to work as expected:

```
$~/Downloads/kubent -c=false --helm2=false --helm3=false -f ingress-v1beta1.yaml
11:27AM INF >>> Kube No Trouble `kubent` <<<
11:27AM INF version 0.5.1 (git sha a762ff3c6b5622650b86dc982652843cc2bd123c)
11:27AM INF Initializing collectors and retrieving data
11:27AM INF Retrieved 1 resources from collector name=File
11:27AM INF Loaded ruleset name=custom.rego.tmpl
11:27AM INF Loaded ruleset name=deprecated-1-16.rego
11:27AM INF Loaded ruleset name=deprecated-1-22.rego
11:27AM INF Loaded ruleset name=deprecated-1-25.rego
__________________________________________________________________________________________
>>> Deprecated APIs removed in 1.22 <<<
------------------------------------------------------------------------------------------
KIND      NAMESPACE     NAME           API_VERSION          REPLACE_WITH (SINCE)
Ingress   <undefined>   test-ingress   extensions/v1beta1   networking.k8s.io/v1 (1.19.0)
```

Can you elaborate a bit more on what exactly are you trying and is not working? If we're talking about in-cluster resource, can you share the full manifest (`kubectl get ... -o yaml`)? Cheers 👍

mausch:
Hi, thanks for checking back.
Just noticed that if I download the manifest and pass it to kubent explicitly it does recognise it as deprecated.
But when I let kubent scan in-cluster resources it doesn't report it.

[...]

PR #98, opened 2021-02-03
fix: Deprecated networking.k8s.io/v1beta1

dja-fr:
Fix https://github.com/doitintl/kube-no-trouble/issues/94


sebastiaankok:
Both Ingress APIs networking.k8s.io/v1beta1 and extensions/v1beta1 will be removed in 1.22 and should get replaced with networking.k8s.io/v1 which is available in Kubernetes 1.19

I've created a new PR for this ->  #99

stepanstipl:
Thanks for this 👍 , and sorry for taking a while to get back to this.  This has now been included in PR #145 that covers a wider set of resources, please have a look there if you have any comments/suggestions.

I'm going to close this one in favour of #145.

PR #99, opened 2021-02-04
fix: Deprecated networking.k8s.io/v1beta1 #2

### LINKED ISSUE #94: networking.k8s.io/v1beta1 should be flagged deprecated also

sebastiaankok:
fix #94 

Both Ingress APIs networking.k8s.io/v1beta1 and extensions/v1beta1 will be removed in 1.22 and should get replaced with  networking.k8s.io/v1 which is available in Kubernetes 1.19 

Ran the build and tested kubent on different k8s clusters. 

```
./kubent  
12:07PM INF >>> Kube No Trouble `kubent` <<<
12:07PM INF version 4merge (git sha 5b16c0624cb77dd7dab69d9a114cd00ce6c82dd6)
12:07PM INF Initializing collectors and retrieving data
12:07PM INF Retrieved 51 resources from collector name=Cluster
12:07PM INF Retrieved 11 resources from collector name="Helm v2"
12:07PM INF Retrieved 63 resources from collector name="Helm v3"
12:07PM INF Loaded ruleset name=deprecated-1-16.rego
12:07PM INF Loaded ruleset name=deprecated-1-22.rego
>>> Deprecated APIs removed in 1.22 <<<
KIND      NAMESPACE            NAME                             API_VERSION                 REPLACE_WITH (SINCE)
Ingress   default              nginx-static-static              networking.k8s.io/v1beta1   networking.k8s.io/v1 (1.19+)
Ingress   kuberhealthy         kuberhealthy                     extensions/v1beta1          networking.k8s.io/v1 (1.19+)
```


stepanstipl:
Thanks @sebastiaankok for this PR 👍 This totally makes sense, but I would like to provide value to users running older K8s versions too (like GKE 1.16/1.17 is in stable/regular channels now).

So I'm thinking how to better handle this case in general - let's say someone runs this with 1.17 - there's no point showing a warning about an upgrade to resources available only in 1.19, but still showing an upgrade to a resource that is available ( `networking.k8s.io/v1beta1` in this case) would be valuable.
 
So for that reason, I would prefer not to replace, but to add this. But this brings up a question of dealing with duplicates, I.e. no point showing advice to upgrade both to `networking.k8s.io/v1beta1` and `networking.k8s.io/v1`.
 
I'm thinking the best way would be something like:
 - Detect all deprecations (i.e. there would be a duplicate in this case - `networking.k8s.io/v1beta1` -> `networking.k8s.io/v1` and `extensions/v1beta1` -> `networking.k8s.io/v1beta1`
 - Filter and show only the newest that is relevant for the cluster that user is running (if known, otherwise show only latest)
 
Some possible issues - duplication in the rego rules (i.e. one for `v1beta1` and one for `v1`, and all would need to cover all previous version of the resource), edge-cases?
 
An alternative might be extending/changing the data format to allow tracking "evolution" of resource, but this would be. a bigger change, also requiring restructuring how resources are organized currently.
 
 
So this is my thinking, any input on this would be appreciated  - would the above work well for you? Any ideas/comments?
 Thanks

sebastiaankok:
Hi Stepan, I hope that I can provide some feedback for you on this. 
When considering that the tool currently reports the user removed API's in v1.22, I think it's best to only show the `networking.k8s.io/v1` api - since both `networking.k8s.io/v1beta1` and `extensions/v1beta1` are still available in v1.19 

I understand the value of adding info about deprecation of v1beta1, but from a user perspective it would make more sense if this was only a deprecation notice/warning. I think "removed" and "deprecated" api's should be handled differently.

stepanstipl:
Hi @sebastiaankok thanks, and sorry for taking ages to get back to this. I think you're right, and until we introduce some more complex mechanism to handle this better, this change totally makes sense and provides better value to the user.

This has now been included in the wider #145 PR, please have a look there if it looks ok to you or if you have any suggestions. I'll close this one, for now.

PR #109, opened 2021-02-26
ref: Replace statik with Go embed

### LINKED ISSUE #107: Use go embed instead of static
Want to avoid using generate and load the rules into the go binary at either build time or run time.

See https://tip.golang.org/pkg/embed/

over static

david-doit-intl:
Closes #107 


stepanstipl:
Ok, I like getting rid of statik, looks simpler and certainly nice to use Go-native feature now that it's available 🎉 🎉 , thanks @david-doit-intl!

2 minor things:
- Can we split this in 2 PRs - one for Go 1.16 upgrade (or at least commits, but changelog is generated based on PRs), one for getting rid of statik
- Can we keep the rules in top-level directory? I would a) prefer to have those there and b) not to change the path. Or is this a limitation of the embed?

stepanstipl:
Just noticed there's already #101 by bot, perhaps we can move the whole Go 1.16 upgrade there.

david-doit-intl:
@stepanstipl yeah lets get 1.16 in first with the bot, then I can merge :). Will try to move rules back to top level, had some issues but will try again.

david-doit-intl:
@stepanstipl updated to isolate rules to its own package pretty happy now.

cmd/kubent/main.go
```diff
@@ -96,7 +97,12 @@ func main() {
 	initCollectors := initCollectors(config)
 	collectors := getCollectors(initCollectors)
 
-	judge, err := judge.NewRegoJudge(&judge.RegoOpts{})
+	loadedRules, err := rules.FetchRegoRules()
```
david-doit-intl, line 100: We should make some mock rules at some point.

pkg/judge/rego_test.go
```diff
@@ -64,7 +65,12 @@ func TestEvalRules(t *testing.T) {
 				manifests = append(manifests, manifest)
 			}
 
-			judge, err := NewRegoJudge(&RegoOpts{})
+			loadedRules, err := rules.FetchRegoRules()
```
david-doit-intl, line 68: We should make some mock rules in a later PR

stepanstipl (approved):
Very nice 👍 , I really like this last iteration 🚀

PR #110, opened 2021-02-26
feat: Windows release

### LINKED ISSUE #108: Using Kube-no-trouble in windows OS
Good morning, 

There're some way to use kubent into windows OS system?
unfortunadle, I use windows to work with kubernetes clusters and we want to upgrade our clusters to 1.16 and need some way to use kubent to locate used deprecated API versions in 1.16.

Good job and thanks!, regards.

david-doit-intl:
Closes #108

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -26,8 +26,8 @@ jobs:
       run: |
         scripts/alpine-setup.sh
         make test
-  build:
-    name: Build
+  build-darwin:
+    name: build darwin
     runs-on: ubuntu-latest
     container: golang:1.16.2-alpine3.12
     steps:
@@ -39,6 +39,26 @@ jobs:
         run: |
           scripts/alpine-setup.sh
           GOOS=darwin GOARCH=amd64 make all
+        shell: sh
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Archive release artifacts
+        uses: actions/upload-artifact@v1
+        with:
+          name: release-artifacts-darwin
+          path: release-artifacts
+  build:
+    name: build linux
+    runs-on: ubuntu-latest
+    container: golang:1.16.1-alpine3.12
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+      - name: Build
+        run: |
+          scripts/alpine-setup.sh
           GOOS=linux GOARCH=amd64 make all
           make changelog
         shell: sh
@@ -49,15 +69,45 @@ jobs:
         with:
           name: release-artifacts
           path: release-artifacts
+  build-windows:
+    name: build windows
+    runs-on: ubuntu-latest
+    container: golang:1.16.1-alpine3.12
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+      - name: Build
+        run: |
+          scripts/alpine-setup.sh
+          GOOS=windows GOARCH=amd64 make all
+        shell: sh
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Archive release artifacts
+        uses: actions/upload-artifact@v1
+        with:
+          name: release-artifacts-windows
+          path: release-artifacts
   release:
     name: Release
-    needs: build
+    needs:
+      [test, pre-commit, build, build-darwin, build-windows]
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
     steps:
       - uses: actions/download-artifact@v1
         with:
           name: release-artifacts
+          path: release-artifacts
+      - uses: actions/download-artifact@v1
+        with:
+          name: release-artifacts-darwin
+          path: release-artifacts
+      - uses: actions/download-artifact@v1
+        with:
+          name: release-artifacts-windows
           path: release-artifacts
       - name: Get the tag
         id: get_tag
@@ -80,7 +130,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./release-artifacts/kubent-${{ steps.get_tag.outputs.git_tag }}-darwin-amd64.tar.gz
+          asset_path: ./release-artifacts-darwin/kubent-${{ steps.get_tag.outputs.git_tag }}-darwin-amd64.tar.gz
           asset_name: kubent-${{ steps.get_tag.outputs.git_tag }}-darwin-amd64.tar.gz
           asset_content_type: application/tar+gzip
       - name: Upload Release Asset - Linux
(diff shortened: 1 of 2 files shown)


stepanstipl:
Nice 👍 , seems the cross-compile works, and as it already works for mac/linux fine, I would kind of hope that it will work for Win too, would be nice to test though - did you manage to?

And I think it will need corresponding `- name: Upload Release Asset - Windows` step added to `.github/workflows/main.yaml` to actually upload the windows binary.

david-doit-intl:
@stepanstipl yeah will add just wanted to see if cross compile worked

david-doit-intl:
@stepanstipl added and removed WIP

stepanstipl:
Cool 👍 I think this looks good, I would just like to do a test on a real windows machine to confirm it actually works before merging this (seen various issues with CGO and dns and stuff historically around K8s go-client - saying that I believe it should not be a problem anymore).

david-doit-intl:
@stepanstipl agreed I don't have a windows machine

gerardgorrion:
Hi all, there're some update of this feature? shows "Merging is blocked". Thanks!

david-doit-intl:
@gerardgorrion apologies, we have been busy. Have you had a chance to try the binary from this build?

stepanstipl:
I did a quick test on Win and found 2 issues:

1) I think the binary has to have `.exe` extension, otherwise I was not able to convince the system to recognize it as executable *(funnily enough once it has the `.exe`, you can invoke it both by `kubent-windows-amd64` and `kubent-windows-amd64.exe` names... both will execute the `.exe` binary)*
-> we should name the download artefact that way.

2) The coloured output is messed up - probably we should disable it for windows, I assume it won't recognize the same terminal control characters. Happy to handle this in a separate issue, as it's not affecting core functionality, just looks ugly ↓

Otherwise hooray 🎉  - seems to work and collect resources!

Screenshot 2021-03-11 at 12 34 01

david-doit-intl:
Ok, this is now a .exe

david-doit-intl:
@stepanstipl this one is all done now just awaiting your review

stepanstipl:
I'm afraid this didn't quite work:

```bash
$ GOOS=windows GOARCH=amd64 gmake all
...
$ cd release-artifacts
$ tar -xvzf kubent-dev-windows-amd64.tar.gz
x kubent
$ ls
kubent                          kubent-dev-windows-amd64.tar.gz
```

Seems like the binary in the archive is still called just `kubent`.

david-doit-intl:
@stepanstipl my bad it looks like it just packing the wrong one, will fix.
```
ls bin
kubent-linux-amd64  kubent-windows-amd64.exe  packed
```

david-doit-intl:
I had to add another nasty if statement tho

.github/workflows/main.yaml
```diff
@@ -49,16 +69,46 @@ jobs:
...
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
     steps:
       - uses: actions/download-artifact@v1
         with:
           name: release-artifacts
           path: release-artifacts
+      - uses: actions/download-artifact@v1
+        with:
+          name: release-artifacts-darwin
+          path: release-artifacts
```
stepanstipl, line 107: I think this will put the artifacts in the `release-artifacts` directory, i.e. the next upload release asset step will fail.

.github/workflows/main.yaml
```diff
@@ -49,16 +69,46 @@ jobs:
...
+      - uses: actions/download-artifact@v1
+        with:
+          name: release-artifacts-windows
+          path: release-artifacts
```
stepanstipl, line 111: Same problem.

.github/workflows/main.yaml
```diff
@@ -26,8 +26,8 @@ jobs:
       run: |
         scripts/alpine-setup.sh
         make test
-  build:
-    name: Build
+  build-darwin:
```
stepanstipl, line 29: Would this be a good fit for using the _matrix_ functionality to avoid some repetition?
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

stepanstipl (commented):
Good stuff, some suggestions:

I would suggest introducing a new variable to avoid 3x repeated if condition (and use make logic):
```make
ifeq (windows,$(GOOS))
  BIN_RELEASE_SUFFIX ?= .exe
endif
BIN_RELEASE_SUFFIX ?=
```

I think it's a bit cleaner. Then we can just change the name in the transform arg:
```shell
$(TAR) -cvz --transform 's,$(PACKED_DIR)/$(*)-$(BIN_ARCH),$(*)$(BIN_RELEASE_SUFFIX),gi' -f "$@" "$<"
```

This has the benefit/disadvantage - depending how you look at it, as compared to the current implementation, of the binary name being consistent, and only appending the `.exe` when we build the archive. WDYT?

.github/workflows/main.yaml
david-doit-intl: yes working on it

Makefile
```diff
@@ -32,11 +37,11 @@ CMD_DIR ?= cmd
 RELEASE_DIR ?= release-artifacts
 PACKED_DIR ?= $(BIN_DIR)/packed
 CMDS ?= $(shell ls $(CMD_DIR))
-BINS ?= $(addsuffix -$(GOOS)-$(GOARCH),$(addprefix $(BIN_DIR)/,$(CMDS)))
+BINS ?= $(addsuffix -$(GOOS)-$(GOARCH)$(BIN_RELEASE_SUFFIX),$(addprefix $(BIN_DIR)/,$(CMDS)))
```
stepanstipl, line 40: I think this is not needed anymore.

Makefile
```diff
@@ -32,11 +37,11 @@ CMD_DIR ?= cmd
...
 CHANGELOG ?= changelog.md
 
-BIN_ARCH ?= $(GOOS)-$(GOARCH)
-RELEASE_SUFFIX ?= $(GIT_REF)-$(BIN_ARCH).tar.gz
+BIN_ARCH ?= $(GOOS)-$(GOARCH)$(BIN_RELEASE_SUFFIX)
```
stepanstipl, line 43: Same - I think this is not needed anymore.

.github/workflows/main.yaml
```diff
@@ -27,9 +27,12 @@ jobs:
         scripts/alpine-setup.sh
         make test
   build:
-    name: Build
+    name: build
     runs-on: ubuntu-latest
-    container: golang:1.16.2-alpine3.12
+    container: golang:1.16.1-alpine3.12
```
stepanstipl, line 32: I merged this yesterday, can you please rebase so that we're not reverting this change.

.github/workflows/main.yaml
```diff
@@ -38,27 +41,28 @@ jobs:
       - name: Build
         run: |
           scripts/alpine-setup.sh
-          GOOS=darwin GOARCH=amd64 make all
-          GOOS=linux GOARCH=amd64 make all
+          GOOS=${{ matrix.os }} GOARCH=amd64 make all
           make changelog
+          mv release-artifacts release-artifacts-${{ matrix.os }}
```
stepanstipl, line 46: I think this is not needed if the path below will use just `release-artifacts`, ... or I might be lost in how GH actions work here :)

.github/workflows/main.yaml
```diff
@@ -38,27 +41,28 @@ jobs:
...
       - name: Archive release artifacts
         uses: actions/upload-artifact@v1
         with:
-          name: release-artifacts
-          path: release-artifacts
-  release:
-    name: Release
-    needs: build
+          name: release-artifacts-${{ matrix.os }}
+          path: release-artifacts-${{ matrix.os }}
```
stepanstipl, line 54: Related to the above - if we remove the var here, I think we don't need the `mv` above

stepanstipl (commented):
Some minor bits to clean up, but I think it looks good and it's pretty much there! 🎉 

Super nice that you used the matrix functionality also for uploading the artefacts! 👍 👍

stepanstipl (approved):
Looks GOOD!

PR #111, opened 2021-02-28
feat: Add support for custom resources

### LINKED ISSUE #44: Feature request: Customizing ruleset
It would be good to have functionalities like checking objects deployed using some specific Helm version i.e. Helm v2 which is gonna be deprecated soon and report that also. In short, allowing to incorporate custom rules also on top of K8s deprecation rules.

stepanstipl:
~*WIP: Still needs:*~
- [x] ~docs update~
- [x] ~(at least) some test coverage~
- [x] ~test to test the custom rule~
- [x] ~squash before merge~

---

This PR adds support for detecting and flagging custom resources.
```
kubent -a "ManagedCertificate.v1.networking.gke.io"
```

The `-a` flag, or it's full form `--additional-kind`, can be used multiple times.
The expected format is full `Kind.version.group.com` form - e.g. `ManagedCertificate.v1.networking.gke.io`.

Closes #44

### DIFF
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth/azure"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -97,7 +98,15 @@ func main() {
 	initCollectors := initCollectors(config)
 	collectors := getCollectors(initCollectors)
 
-	loadedRules, err := rules.FetchRegoRules()
+	// this could probably use some error checking in future, but
+	// schema.ParseKindArg does not return any error
+	var additionalKinds []schema.GroupVersionKind
+	for _, ar := range config.AdditionalKinds {
+		gvr, _ := schema.ParseKindArg(ar)
+		additionalKinds = append(additionalKinds, *gvr)
+	}
+
+	loadedRules, err := rules.FetchRegoRules(additionalKinds)
 	if err != nil {
 		log.Fatal().Err(err).Str("name", "Rules").Msg("Failed to load rules")
 	}
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -55,7 +55,7 @@ func storeCollector(collector collector.Collector, err error, collectors []colle
 func initCollectors(config *config.Config) []collector.Collector {
 	collectors := []collector.Collector{}
 	if config.Cluster {
-		collector, err := collector.NewClusterCollector(&collector.ClusterOpts{Kubeconfig: config.Kubeconfig})
+		collector, err := collector.NewClusterCollector(&collector.ClusterOpts{Kubeconfig: config.Kubeconfig}, config.AdditionalKinds)
 		collectors = storeCollector(collector, err, collectors)
 	}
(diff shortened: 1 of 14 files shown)


david-doit-intl:
Could we add a test where additional is not  []string{}?

pkg/judge/rego.go
```diff
@@ -19,17 +24,24 @@ type RegoJudge struct {
...
+
 	r := rego.New(
 		rego.Query("data[_].main"),
 	)
 
 	statikFS, err := fs.New()
 
-	fs.Walk(statikFS, "/",
+	err = fs.Walk(statikFS, "/",
 		func(path string, info os.FileInfo, err error) error {
```
david-doit-intl, line 43: this is too large now, please make it a separate go function

david-doit-intl (changes requested):
Some minor cleanup, but looks good

stepanstipl:
I think this is ready for another review. I did rebase and refactor to accommodate the changes introduced with the use of the embed FS, addressed the too long function mentioned above and added unit tests.

What still needs to be done before merge:
- The docs still need to be updated
- and I plan to add at least one "functional" test to cover the checking for additional resource

pkg/judge/rego.go
stepanstipl: This should be addressed now.

stepanstipl:
@david-doit-intl sorry, I can't see where you commented as I noticed the failing test and was fixing the fmt error. If it was that, it's fixed, if smth. else, can you please comment again? :)

pkg/rules/rules_test.go
```diff
@@ -26,3 +29,93 @@ func TestFetchRules(t *testing.T) {
 		}
 	}
 }
+
+func TestFetchRulesWithAdditionalResources(t *testing.T) {
+	var expected []string
+	root := "rego/"
```
david-doit-intl, line 35: use the const

pkg/rules/rules_test.go
```diff
@@ -26,3 +29,93 @@ func TestFetchRules(t *testing.T) {
...
+	err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if info.Name() != "rego" {
```
david-doit-intl, line 37: here too

cmd/kubent/main.go
```diff
@@ -79,13 +80,14 @@ func initCollectors(config *config.Config) []collector.Collector {
...
+	zerolog.SetGlobalLevel(zerolog.InfoLevel)
+
 	config, err := config.NewFromFlags()
 	if err != nil {
 		log.Fatal().Err(err).Msg("failed to parse config flags")
 	}
 
-	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
-	zerolog.SetGlobalLevel(zerolog.InfoLevel)
 	if config.Debug {
```
david-doit-intl, line 91: might want to move that too!

pkg/rules/rules_test.go
stepanstipl: Done.

pkg/rules/rules_test.go
stepanstipl: Done.

stepanstipl:
Added the "integration" test to test the actual custom rule functionality. It's in a separate package `test`, as it's using other packages and can't sit under rules (cause cyclic imports). I think it can serve as a starting point for implementing tests for the individual rules.

I think it's ready for final review @david-doit-intl

david-doit-intl (approved):

PR #121, opened 2021-03-31
feat: jsonl output and quiet stdout

leepa:
To support better automated usage.

New options:

 * -q will quiet the stderr log output
 * -o jsonl will, instead out outputting a list, output one infraction
   per line. This is good if you're loading the data into data
   warehouses or other reporting tools.

### DIFF
--- a/pkg/printer/json.go
+++ b/pkg/printer/json.go
@@ -9,27 +9,50 @@ import (
 )
 
 type JSONPrinter struct {
+	opts *JSONOpts
 }
 
 type JSONOpts struct {
+	// OnePerLine enables JSONL support
+	OnePerLine bool
 }
 
 func NewJSONPrinter(opts *JSONOpts) (*JSONPrinter, error) {
-	printer := &JSONPrinter{}
+	printer := &JSONPrinter{opts: opts}
 
 	return printer, nil
 }
 
 func (c *JSONPrinter) Print(results []judge.Result) error {
-
 	buffer := new(bytes.Buffer)
-	encoder := json.NewEncoder(buffer)
-	encoder.SetIndent("", "\t")
 
-	err := encoder.Encode(results)
-	if err != nil {
-		return err
+	if !c.opts.OnePerLine {
+		encoder := json.NewEncoder(buffer)
+		encoder.SetIndent("", "\t")
+
+		err := encoder.Encode(results)
+		if err != nil {
+			return err
+		}
+	} else {
+		for _, entry := range results {
+			b, err := json.Marshal(entry)
+			if err != nil {
+				return err
+			}
+
+			_, err = buffer.Write(b)
+			if err != nil {
+				return err
+			}
+
+			err = buffer.WriteByte('\n')
+			if err != nil {
+				return err
+			}
+		}
 	}
+
 	fmt.Printf("%s", buffer.String())
 
 	return nil
--- a/pkg/printer/json_test.go
+++ b/pkg/printer/json_test.go
@@ -1,0 +1,114 @@
+package printer
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"log"
+	"os"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/doitintl/kube-no-trouble/pkg/judge"
+)
+
+var findingsJsonTesting []judge.Result = []judge.Result{
+	{
+		Name:        "testName1",
+		Kind:        "testKind1",
+		Namespace:   "testNamespace1",
+		ApiVersion:  "v1",
+		RuleSet:     "testRuleset1",
+		ReplaceWith: "testReplaceWith1",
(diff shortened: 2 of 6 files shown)


pkg/printer/json_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func TestNewJsonPrinter(t *testing.T) {
+	x, _ := NewJSONPrinter(&JSONOpts{})
+	if x.opts.OnePerLine != false {
+		t.Errorf("OnePerLine should default to false")
+	}
+
+	x, _ = NewJSONPrinter(&JSONOpts{OnePerLine: true})
+	if x.opts.OnePerLine != true {
+		t.Errorf("OnePerLine should default to false")
```
david-doit-intl, line 75: OnePerLine should be set to true

pkg/printer/json_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+func TestPrintJson(t *testing.T) {
+	x, _ := NewJSONPrinter(&JSONOpts{OnePerLine: false})
```
david-doit-intl, line 80: x -> jsonPrinter

pkg/printer/json.go
```diff
@@ -9,27 +9,50 @@ import (
...
+		encoder := json.NewEncoder(buffer)
+		encoder.SetIndent("", "\t")
+
+		err := encoder.Encode(results)
+		if err != nil {
+			return err
+		}
+	} else {
+		for _, entry := range results {
+			b, err := json.Marshal(entry)
```
david-doit-intl, line 39: b-> bytes?

pkg/printer/json_test.go
leepa: Good catch

david-doit-intl (changes requested):
LGTM aside from some little changes @stepanstipl WDYT

stepanstipl:
Thanks @leepa for contributing! Before looking at the implementation details, I would suggest splitting these into 2 PRs, as these seem to be unrelated changes, so as not to mix the conversation. But TL;DR: for both of those I would like to discuss/see an example of use cases, more specifically:

- `-q` - is there any benefit/use case when same can't be achieved by ignoring stderr?
- `-o jsonl` - kind of same, I'm a bit concerned about introducing a somewhat arbitrary output format, but I would like to understand the use case first. Can we also get an example of output? It seems to me that the current output won't be valid json?

stepanstipl:
> Tools like Prowler and c7n already support jsonl. In Prowler's case it's the default for json output and in c7n, it's how it then puts things into SQS for notification. Far from arbitrary, it makes for easier importing into a data warehouse or documents in ElasticSearch. It's also the format that SecurityHub uses.

I'm not well familiar with these tools, and not sure about the use case - how would you use this `kubent` with Prowler? If the use case is putting `kubent` output into SQS, can you expand on that? 

Re. _AWS SecurityHub_ - it seems to require what is called _[AWS Security Finding Format (ASFF)](https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format-syntax.html)_ format. Is the use case actually importing into AWS SecurityHub? If so, can you share any links to help me understand how SH requires single-lines json and how this would be integrated together?

I've found some info about [JSON lines](https://jsonlines.org/). As I understand it this format is intended for a stream, one record at a time, processing. I'm not sure it's a fit, as that is not something supported by `kubent`.

If the use case is running from a command line, perhaps the easy way is to use smth like _jq_ - you can easily run the following to convert:
```bash
$ kubent -o json | jq -c '.[]'
```

If the above doesn't work, can you elaborate a bit on the specific use case? You mentioned a couple of tools, but I'm still not very clear on how is this intended to be used.


> -q is also something both tools have - for the same reason. At the very least, it would be good to be able to adjust the log level properly - which is probably nicer?

Yep, I don't think the `-q` is the way to go here - it doesn't really silence _stdout,_ nor _stderr_ - just changes log level in the current implementation. On the other hand, adding an option to adjust the log level sounds like a good idea 👍.  This brings me back to understanding the use case... would custom log level work for you? For example simple mapping to zerolog's [levels](https://github.com/rs/zerolog#leveled-logging)?

> Not sure we need to split this up.

It seems like these features are orthogonal. It's easier to review, discuss and move forward smaller chunks of work (Not to mention how release notes are generated and split to at least separate commits will be required).

stepanstipl:
Configurable log level has been implemented in #127, in particular with `disabled` option to cover this use case. Thans for the suggestion 👍 .

As to JSON lines output format - I haven't heard more about the use case, but as for the command line use the simple way is to use `jq` (`kubent -o json | jq -c '.[]'`) to achieve the same, and gives the user the additional flexibility to transform to any format they're after. Additionally, as we don't support streaming, there doesn't seem to be any benefit in implementing this format inside this tool, and currently, there's no plan to do so.

PR #134, opened 2021-04-20
ref: Printer Updates

david-doit-intl:
Was very WIP

### DIFF
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -119,18 +119,12 @@ func main() {
 		log.Fatal().Err(err).Str("name", "Rego").Msg("Failed to evaluate input")
 	}
 
-	var prnt printer.Printer
-	switch config.Output {
-	case "json":
-		prnt, err = printer.NewJSONPrinter(&printer.JSONOpts{})
-	default:
-		prnt, err = printer.NewTextPrinter(&printer.TextOpts{})
-	}
+	printer, err := printer.NewPrinter(config.Output)
 	if err != nil {
-		log.Fatal().Err(err).Str("name", config.Output).Msg("Failed to initialize output printer")
+		log.Fatal().Err(err).Msg("Failed to create printer")
 	}
 
-	err = prnt.Print(results)
+	err = printer.Print(results)
 	if err != nil {
 		log.Fatal().Err(err).Msg("Failed to print results")
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -13,25 +13,6 @@ import (
 	"k8s.io/client-go/util/homedir"
 )
 
-type ZeroLogLevel int8
-
-func (l ZeroLogLevel) String() string {
-	return zerolog.Level(l).String()
-}
-
-func (l *ZeroLogLevel) Set(s string) error {
-	level, err := zerolog.ParseLevel(s)
-	if err != nil {
-		return err
-	}
-	*l = ZeroLogLevel(level)
-	return nil
-}
-
-func (l ZeroLogLevel) Type() string {
-	return "string"
-}
-
 type Config struct {
 	AdditionalKinds []string
 	Cluster         bool
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"unicode"
 
+	"github.com/doitintl/kube-no-trouble/pkg/printer"
 	"github.com/rs/zerolog"
 	flag "github.com/spf13/pflag"
 	"k8s.io/client-go/util/homedir"
@@ -44,7 +45,9 @@ type Config struct {
 }
 
 func NewFromFlags() (*Config, error) {
-	config := Config{LogLevel: ZeroLogLevel(zerolog.InfoLevel)}
+	config := Config{
+		LogLevel: ZeroLogLevel(zerolog.InfoLevel),
+	}
 
 	home := homedir.HomeDir()
 	flag.StringSliceVarP(&config.AdditionalKinds, "additional-kind", "a", []string{}, "additional kinds of resources to report in Kind.version.group.com format")
@@ -58,6 +61,10 @@ func NewFromFlags() (*Config, error) {
 	flag.VarP(&config.LogLevel, "log-level", "l", "set log level (trace, debug, info, warn, error, fatal, panic, disabled)")
 
 	flag.Parse()
+
+	if _, err := printer.ParsePrinter(config.Output); err != nil {
+		return nil, fmt.Errorf("failed to validate argument output: %w", err)
+	}
 
 	if err := validateAdditionalResources(config.AdditionalKinds); err != nil {
 		return nil, fmt.Errorf("failed to validate arguments: %w", err)
--- a/pkg/printer/json.go
+++ b/pkg/printer/json.go
@@ -8,29 +8,31 @@ import (
 	"github.com/doitintl/kube-no-trouble/pkg/judge"
 )
 
-type JSONPrinter struct {
+type jsonPrinter struct {
 }
 
-type JSONOpts struct {
+func newJSONPrinter() Printer {
+	return &jsonPrinter{}
 }
 
-func NewJSONPrinter(opts *JSONOpts) (*JSONPrinter, error) {
-	printer := &JSONPrinter{}
-
-	return printer, nil
-}
-
-func (c *JSONPrinter) Print(results []judge.Result) error {
-
+func (c *jsonPrinter) populateOutput(results []judge.Result) (string, error) {
 	buffer := new(bytes.Buffer)
 	encoder := json.NewEncoder(buffer)
 	encoder.SetIndent("", "\t")
 
 	err := encoder.Encode(results)
 	if err != nil {
+		return "", err
+	}
+	return buffer.String(), nil
+}
+
+func (c *jsonPrinter) Print(results []judge.Result) error {
+	output, err := c.populateOutput(results)
+	if err != nil {
 		return err
 	}
-	fmt.Printf("%s", buffer.String())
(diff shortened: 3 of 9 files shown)


stepanstipl:
First, can we pls. rebase/merge - currently PR is showing diffs for changes that have landed in master _(I've tried to change base back & forth ^^, but that didn't seem to help)_.

Now I would like to keep the conversation about the refactoring and _jsonl_ separate, otherwise, it will get mixed together. So my proposal is move `jsonl` related changes to another PR (they would need to be in separate commits anyway), and let's discuss the refactor here.

Re. refactor: I like the general direction of moving the logic away from `main` 👍 .

What I'm not sure about is introducing the new type `PrinterType`:
```go
type PrinterType struct {
	Printer printer.Printer
	name    string
}
```

It doesn't seem to bring much value, and I think we can get away with using the `Printer` type, that already exists, directly in the config to hold the pointer to the `Printer` instance. And not introduce this additional abstraction layer. But maybe I'm missing smth.?

`+` thanks for the added tests! 🎉

david-doit-intl:
@stepanstipl I am pretty sure this pr is too date. Not sure what exactly you are seeing. I am happy to move the jsonl out no worries!

david-doit-intl:
@stepanstipl printer is an interface not a concrete type so we need something to hold it which is concrete, additionally the name stores what printer you selected, I also plan to add args to the struct at a later date, so the struct seems the most reasonable

david-doit-intl:
Additionally I want to make the printer type less tightly coupled to the printer module

stepanstipl:
> @stepanstipl I am pretty sure this pr is too date. Not sure what exactly you are seeing.

There are some log level changes in the diff - https://github.com/doitintl/kube-no-trouble/pull/134/files#diff-e1976a6dedca1ef99072ad60854c3e58bd53e68b07dc7a04cb0102e08a115623

> I am happy to move the jsonl out no worries!

👍 , thanks.

> @stepanstipl printer is an interface not a concrete type so we need something to hold it which is concrete, additionally the name stores what printer you selected, I also plan to add args to the struct at a later date, so the struct seems the most reasonable

So in the current implementation, the config holds `PrinterType`:
```go
		Output: config.PrinterType{
			Printer: &printer.TextPrinter{},
		},
```
and this, in turn, holds pointer to an instance of some type satisfying the `Printer` interface:
```go
type PrinterType struct {
	Printer printer.Printer
	name    string
}
```
So it's `Config.Output -----> PrinterType -----> *Printer`.

We can do just `Config.Output ----->*Printer`.


The name can become property/method of the Printer interface

david-doit-intl:
@stepanstipl no, thats the who point. Printer is a interface not a concrete type.

david-doit-intl:
though I am happy to make name part of the interface

david-doit-intl:
@stepanstipl have a look now

pkg/printer/printer.go
```diff
@@ -4,6 +4,15 @@ import (
 	"github.com/doitintl/kube-no-trouble/pkg/judge"
 )
 
+var Printers = map[string]Printer{
```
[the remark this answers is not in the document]
stepanstipl: Thinking about this more, maybe the solution is to have this map, but instead of storing the printers, we could store relevant functions to create those printers, and call once needed?

pkg/config/config.go
```diff
@@ -59,6 +43,10 @@ func NewFromFlags() (*Config, error) {
 
 	flag.Parse()
 
+	if _, exists := printer.Printers[config.Output]; !exists {
```
stepanstipl, line 46: Maybe we can use some function (align with `ParsePrinter`?)  instead of accessing the `.Printers` map directly. Making me think the `.Printers` could even be private, otherwise this is very tightly coupled.

stepanstipl (commented):
Some changes around how the printers are created, apart from that I think it's good. I like moving the logic away from the main.

👍 for the buffer change & covering it with test.

Some housekeeping details (can be done later, before final merge):
- probably should be `ref:` commit, as refactoring
- split the zerolog related changes to a separate `ref:` commit

pkg/printer/printer.go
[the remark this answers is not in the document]
david-doit-intl: Yeah thats what I wanted ^^, I couldn't find away to do it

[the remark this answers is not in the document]
david-doit-intl: I wanted:

map[String]func -> Printer

pkg/printer/json.go
```diff
@@ -11,26 +11,24 @@ import (
...
-
-func NewJSONPrinter(opts *JSONOpts) (*JSONPrinter, error) {
-	printer := &JSONPrinter{}
-
-	return printer, nil
-}
-
-func (c *JSONPrinter) Print(results []judge.Result) error {
-
+func (c *JSONPrinter) populateBuffer(results []judge.Result) (string, error) {
```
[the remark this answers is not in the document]
david-doit-intl: hmm, strings are more useable, so I'll rename

pkg/config/config.go
david-doit-intl: perfect! YES love this!

pkg/printer/json.go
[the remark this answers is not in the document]
david-doit-intl: populateOutput

pkg/printer/printer.go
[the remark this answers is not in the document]
david-doit-intl: ok found a way to do it

david-doit-intl:
@stepanstipl much happier now :)

cmd/kubent/main.go
```diff
@@ -119,18 +119,8 @@ func main() {
...
-		prnt, err = printer.NewJSONPrinter(&printer.JSONOpts{})
-	default:
-		prnt, err = printer.NewTextPrinter(&printer.TextOpts{})
-	}
-	if err != nil {
-		log.Fatal().Err(err).Str("name", config.Output).Msg("Failed to initialize output printer")
-	}
-
-	err = prnt.Print(results)
+	printer, _ := printer.ParsePrinter(config.Output)
```
stepanstipl, line 122: We shouldn't be ignoring the error here. Also the variable name is confusing, as this is not a printer, more like `newPrinterFn`.

cmd/kubent/main.go
```diff
@@ -119,18 +119,8 @@ func main() {
...
+	err = printer().Print(results)
```
stepanstipl, line 123: It would be a bit nicer to save the printer, and then call `Print` on it, this feels a bit too much at once + not saving the printer anywhere.

stepanstipl: Thinking about this a bit more, it might be better to offload this to a separate fn in the printer package (not necessary, just to keep main code clean) and combine it with the previous, e.g.:
```go
printer, err := printer.newPrinter(config.Output)
if err...
...
err = printer.Print(results)
```

pkg/printer/printer.go
```diff
@@ -1,9 +1,24 @@
...
+}
+
 type Printer interface {
 	Print([]judge.Result) error
 }
+
+func ParsePrinter(choice string) (func() Printer, error) {
+	printer, exists := printers[choice]
+	if !exists {
+		return nil, fmt.Errorf("failed to validate argument output")
```
stepanstipl, line 21: We should not mention argument in this module
```suggestion
		return nil, fmt.Errorf("unknown printer type %s", choice)
```

pkg/config/config.go
```diff
@@ -59,6 +43,10 @@ func NewFromFlags() (*Config, error) {
 
 	flag.Parse()
 
+	if _, err := printer.ParsePrinter(config.Output); err != nil {
+		return nil, fmt.Errorf("failed to validate argument output")
```
stepanstipl, line 47: I suggest not hiding the error, wrap it instead as below:
```suggestion
		return nil, fmt.Errorf("failed to validate argument output: %w", err)
```

pkg/printer/json_test.go
```diff
@@ -0,0 +1,37 @@
...
+		Kind:        "testKind2",
+		Namespace:   "testNamespace2",
+		ApiVersion:  "v1",
+		RuleSet:     "testRuleset2",
+		ReplaceWith: "testReplaceWith2",
+		Since:       "testSince2",
+	},
+}
+
+func TestPrintJson(t *testing.T) {
```
stepanstipl, line 30: Isn't this testing the `populateOutput`?
```suggestion
func TestJsonPopulateOutput(t *testing.T) {
```

pkg/printer/json_test.go
```diff
@@ -0,0 +1,37 @@
...
+	printer := &JSONPrinter{}
+	_, err := printer.populateOutput(findingsJsonTesting)
+
+	if err != nil {
```
stepanstipl, line 34: I'm wondering, could we test a bit more the expected output, than just that we don't return an error?

stepanstipl (commented):
Yep, much better 👍 . I think it's almost ready, with some minor comments.

david-doit-intl:
Ok all done.

stepanstipl:
Yep, looks good 👍 , can you please do the final housekeeping before the merge & I will merge it:
> Some housekeeping details (can be done later, before final merge):
> 
> probably should be `ref: `commit, as refactoring
> split the zerolog related changes to a separate `ref: ` commit

`+` rebase on current master

david-doit-intl:
> Yep, looks good  , can you please do the final housekeeping before the merge & I will merge it:
> 
> > Some housekeeping details (can be done later, before final merge):
> > probably should be `ref: `commit, as refactoring
> > split the zerolog related changes to a separate `ref: ` commit
> 
> `+` rebase on current master

Yes forgot my bad

stepanstipl (approved):
Cool, thanks 👍  🎉 🍾

Issue #130, opened 2021-04-09
Detect K8s version and suggest only relevant issues


stepanstipl:
We can try to detect a version of the cluster a user is running and suggest only updates relevant for the existing/next minor version, since we have info on since what version are the replacement resources available.

I.e. if user runs 1.15, and is dealing with upgrade to 1.16, no point flagging deprecation in 1.22 that are only available for replacement in 1.18.

stepanstipl:
Some details on how this should work:
- Only detect version if one of the relevant collectors (cluster, Helm v2/3) is enabled _(I'm thinking about extending the collector interface here)_
- Allow version override via flag - for cases when this is used with files in CI
- By default filter findings to the relevant ones only, i.e. if it's possible to upgrade (and we have that info)
  - Q1: Should we filter based on the current cluster version or the next minor - assuming the typical scenario is to use this tool when upgrading clusters. I.e:
    a) `result.Since <= masterMinorVersion`
    b) `result.Since <= masterMinorVersion + 1`
    In b) case you won't be able to fix this until after the upgrade, so seems like a) might make more sense? Downside potentially missing some issues, although this should never happen in practice, as replacement versions are first introduced and old ones removed only after.
  - Q2: What to do about resources being removed, without replacement? I don't think we have this well covered. I'm thinking just warn about those - `Since` will be probably when the resource was marked as deprecated in this case, so we might warn a bit early, but better than late ;).
- Add flag to disable any filtering and display all


@david-doit-intl any thoughts?

david-doit-intl:
1) We should check on the next minor yes :+1: , should have a flag to check up to latest perhaps.
2) We should defo warn about things being removed!

PR #139, opened 2021-04-29
Feat: Detect K8s version

stepanstipl:
This PR adds support for detecting version of K8s server. It's a foundation for being able to filter relevant findings later on.

From #130:
- [x] Only detect version if one of the relevant collectors (cluster, Helm v2/3) is enabled (I'm thinking about extending the collector interface here)
- [x] Allow version override via flag - for cases when this is used with files in CI

WIP, apart from above it's missing:
- [x] Update docs once flag to override version is added
- [x] Tests
- [x] Maybe reduce code duplication - in particular `GetServerVersion` and discovery K8s client creation in `New*Collector` functions (without introducing to much complexity via embedding).
- [x] Needs squash/commits cleanup before merge (please!)

This PR addresses first half of #130.

### DIFF
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -13,7 +13,6 @@ import (
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-
 	_ "k8s.io/client-go/plugin/pkg/client/auth/azure"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
@@ -77,6 +76,16 @@ func initCollectors(config *config.Config) []collector.Collector {
 	return collectors
 }
 
+func getServerVersion(collectors []collector.Collector) (string, error) {
+	for _, c := range collectors {
+		versionCol, ok := c.(collector.VersionCollector)
+		if ok {
+			return versionCol.GetServerVersion()
+		}
+	}
+	return "", nil
+}
+
 func main() {
 	exitCode := EXIT_CODE_FAIL_GENERIC
 
@@ -91,9 +100,18 @@ func main() {
 
 	log.Info().Msg(">>> Kube No Trouble `kubent` <<<")
 	log.Info().Msgf("version %s (git sha %s)", version, gitSha)
-	log.Info().Msg("Initializing collectors and retrieving data")
 
+	log.Info().Msg("Initializing collectors and retrieving data")
 	initCollectors := initCollectors(config)
+
+	if config.TargetVersion == "" {
+		config.TargetVersion, err = getServerVersion(initCollectors)
+		if err != nil {
+			log.Fatal().Err(err).Msg("Failed to detect K8s version")
+		}
+	}
+	log.Info().Msgf("Target K8s version is %s", config.TargetVersion)
+
 	collectors := getCollectors(initCollectors)
 
 	// this could probably use some error checking in future, but
--- a/pkg/collector/collector.go
+++ b/pkg/collector/collector.go
@@ -5,8 +5,18 @@ type Collector interface {
 	Name() string
 }
 
+type VersionCollector interface {
+	GetServerVersion() (string, error)
+}
+
 type commonCollector struct {
 	name string
+}
+
+func newCommonCollector(name string) *commonCollector {
+	return &commonCollector{
+		name: name,
+	}
 }
 
 func (c *commonCollector) Name() string {
--- a/pkg/collector/fake.go
+++ b/pkg/collector/fake.go
@@ -1,0 +1,21 @@
+package collector
+
+const FAKE_VERSION = "1.2.3"
+
+type fakeCollector struct {
+	*commonCollector
+}
+
+func (c *fakeCollector) Get() ([]map[string]interface{}, error) {
+	return []map[string]interface{}{}, nil
+}
+
+func NewFakeCollector() *fakeCollector {
+	return &fakeCollector{
+		commonCollector: newCommonCollector("Fake"),
+	}
+}
+
+func (c *fakeCollector) GetServerVersion() (string, error) {
+	return FAKE_VERSION, nil
+}
--- a/pkg/collector/helm3.go
+++ b/pkg/collector/helm3.go
@@ -2,30 +2,39 @@ package collector
 
 import (
 	"fmt"
-
-	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
 
+	"github.com/ghodss/yaml"
 	"helm.sh/helm/v3/pkg/releaseutil"
 	"helm.sh/helm/v3/pkg/storage"
 	"helm.sh/helm/v3/pkg/storage/driver"
-
-	"github.com/ghodss/yaml"
+	"k8s.io/client-go/discovery"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
 type HelmV3Collector struct {
 	*commonCollector
+	*kubeCollector
 	client       *corev1.CoreV1Client
 	secretsStore *storage.Storage
 	configStore  *storage.Storage
 }
 
 type HelmV3Opts struct {
-	Kubeconfig string
+	Kubeconfig      string
+	DiscoveryClient discovery.DiscoveryInterface
 }
 
 func NewHelmV3Collector(opts *HelmV3Opts) (*HelmV3Collector, error) {
-	collector := &HelmV3Collector{commonCollector: &commonCollector{name: "Helm v3"}}
+	kubeCollector, err := newKubeCollector(opts.Kubeconfig, opts.DiscoveryClient)
+	if err != nil {
+		return nil, err
+	}
+
+	collector := &HelmV3Collector{
+		commonCollector: newCommonCollector("Helm v2"),
+		kubeCollector:   kubeCollector,
+	}
 
 	config, err := clientcmd.BuildConfigFromFlags("", opts.Kubeconfig)
 	if err != nil {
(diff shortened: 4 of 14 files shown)


stepanstipl:
It's WIP, still missing bits as listed above, but would love your feedback @david-doit-intl

cmd/kubent/main.go
```diff
@@ -91,9 +106,18 @@ func main() {
...
-	log.Info().Msg("Initializing collectors and retrieving data")
 
+	log.Info().Msg("Initializing collectors and retrieving data")
 	initCollectors := initCollectors(config)
+
+	version, err := getServerVersion(initCollectors)
+	if err != nil {
+		log.Fatal().Err(err).Msg("Failed to detect K8s version")
+	}
+	if version != nil {
```
stepanstipl, line 117: this should be `else if`, but probably will further change once I add the overriding flag logic.

david-doit-intl: or or

cmd/kubent/main.go
```diff
@@ -77,6 +80,18 @@ func initCollectors(config *config.Config) []collector.Collector {
 	return collectors
 }
 
+func getServerVersion(collectors []collector.Collector) (*k8s_version.Info, error) {
```
david-doit-intl, line 83: this function is a little gross

cmd/kubent/main.go
```diff
@@ -77,6 +80,18 @@ func initCollectors(config *config.Config) []collector.Collector {
...
+	for _, c := range collectors {
+		version, err := c.GetServerVersion()
```
david-doit-intl, line 85: return c.GetServerVersion()

david-doit-intl: you have already checked for error in other function.

pkg/collector/helm3.go
```diff
@@ -43,6 +45,20 @@ func NewHelmV3Collector(opts *HelmV3Opts) (*HelmV3Collector, error) {
 	configDriver := driver.NewConfigMaps(collector.client.ConfigMaps(""))
 	collector.configStore = storage.Init(configDriver)
 
+	if opts.DiscoveryClient == nil {
```
david-doit-intl, line 48: seems like we are repeating ourselves alot here

pkg/collector/helm3.go
```diff
@@ -89,3 +105,11 @@ func (c *HelmV3Collector) Get() ([]map[string]interface{}, error) {
 
 	return results, nil
 }
+
+func (c *HelmV3Collector) GetServerVersion() (*version.Info, error) {
```
david-doit-intl, line 109: and here too!

pkg/errors/errors.go
```diff
@@ -0,0 +1,9 @@
+package errors
```
david-doit-intl, line 1: is this even needed?

david-doit-intl (changes requested):
Hmmm, lots of duplication.

cmd/kubent/main.go
stepanstipl: can you elaborate a bit?

cmd/kubent/main.go
stepanstipl: changed

pkg/collector/helm3.go
stepanstipl: I did a bit of refactoring to get rid of a lot of the duped code. Should be much better now.

pkg/errors/errors.go
stepanstipl: So while this is a single stem in the wind, for now, I was thinking a bit about how to work with errors more conceptually.

If you want to treat errors as consts, you need smth like the:
```go
type Error string
```
(as the Go's inbuilt `errors.errorString` can't be used that way as it's struct).

Now this is pretty convenient, and with improvements of Go's error handling since 1.13 allows you to easily use `errors.Is` to compare errors (incl. wrapped errors - that's what we do with the `fmt.Errorf("... %w", err)`.

I.e. we can easily check if returned error, or any of the wrapped errors inside, is of given type:
```go
if !errors.Is(err, kerrors.ERR_NOT_SUPPORTED) {
   ...
}
```

Finally, instead of doing this in individual packages, I was thinking at least the type should be central.

stepanstipl:
The latest iteration introduces the `collector.VersionCollector`, and then uses
```go
versionCol, ok := c.(collector.VersionCollector)
```
to check if the interface is implemented. The code is ready for another look _(pending missing tests and readme update)._

cmd/kubent/main.go
```diff
@@ -91,9 +100,20 @@ func main() {
...
+	log.Info().Msg("Initializing collectors and retrieving data")
 	initCollectors := initCollectors(config)
+
+	if config.TargetVersion == "" {
+		config.TargetVersion, err = getServerVersion(initCollectors)
+		if err != nil {
+			log.Fatal().Err(err).Msg("Failed to detect K8s version")
+		}
+	}
+	if config.TargetVersion != "" {
```
david-doit-intl, line 113: uhhhh not sure about this, deal with this after you set it

pkg/collector/collector.go
```diff
@@ -5,10 +5,21 @@ type Collector interface {
...
+type VersionCollector interface {
+	GetServerVersion() (string, error)
+}
+
 type commonCollector struct {
 	name string
 }
 
+func newCommonCollector(name string) *commonCollector {
+	col := &commonCollector{
```
david-doit-intl, line 17: do we need col? Seems like:

```
return &commonCollector{
                    name: name
}
```
 would be fine

pkg/collector/kube.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type kubeCollector struct {
+	discoveryClient discovery.DiscoveryInterface
+}
+
+func newKubeCollector(kubeconfig string, discoveryClient discovery.DiscoveryInterface) (*kubeCollector, error) {
```
david-doit-intl, line 13: Better!

pkg/collector/kube.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+	} else {
+		col.discoveryClient = discoveryClient
+	}
+
+	return col, nil
+}
+
+func (c *kubeCollector) GetServerVersion() (string, error) {
+	versn, err := c.discoveryClient.ServerVersion()
```
david-doit-intl, line 34: version!

pkg/config/config.go
```diff
@@ -89,3 +96,19 @@ func validateAdditionalResources(resources []string) error {
 	}
 	return nil
 }
+
+func validateTargetVersion(version string) error {
```
david-doit-intl, line 100: consider a separate file for this? with a test?

cmd/kubent/main.go
```diff
@@ -77,6 +76,16 @@ func initCollectors(config *config.Config) []collector.Collector {
 	return collectors
 }
 
+func getServerVersion(collectors []collector.Collector) (string, error) {
+	for _, c := range collectors {
+		versionCol, ok := c.(collector.VersionCollector)
+		if ok {
```
david-doit-intl, line 82: nice!

david-doit-intl (changes requested):
Much much better!

cmd/kubent/main.go
stepanstipl: Not really, these two if conditions are completely independent. I.e. it can be empty or already set by user when we hit the first `if` block.

And after it, it can still be empty because we didn't detect it. Maybe the wording is misleading? Should probably say smth. like `Target version set to XXX...`, as it doesn't have to be detected and can be coming from command line. Would updating the log message help?

pkg/collector/collector.go
stepanstipl: Done.

pkg/collector/kube.go
stepanstipl: Fingers hurt... too many characters to type... done.

pkg/config/config.go
stepanstipl: Tests are coming 👍 I would keep it here for now, same as the other `validate*`, can rearrange later, maybe move all of those.

cmd/kubent/main.go
stepanstipl: Ok, thinking about this more I just removed the second condition. This will result in the info message being printed even if the version is `""`, but that's probably ok for now. Can adjust later as we add the filtering functionality.

pkg/collector/fake.go
```diff
@@ -0,0 +1,21 @@
+package collector
```
david-doit-intl, line 1: fake_test.go ?

david-doit-intl (changes requested):

david-doit-intl:
@stepanstipl looks good squash some commits and consider renaming fake.go

david-doit-intl:
btw after this pr lands a new release is propably in order

pkg/collector/fake.go
stepanstipl: Agree, but unfortunately, I don't think you can use a  `_test.go` here, as it won't be exported, and therefore not useable from other packages. I have tried it, and a couple of variations to get it working, but if I understand correctly, it's not the way go works. Seems the `*_test.go` files are compiled into the package only when running go test for that package.

stepanstipl:
@david-doit-intl done (the squash, the renaming not gonna work), can you approve before I merge 😉

pkg/collector/fake.go
david-doit-intl: Can we call it mock collector instead?

david-doit-intl (approved):
Fake -> Mock but otherwise :ok_hand:

pkg/collector/fake.go
stepanstipl: Kind of followed K8s naming here - all the resources/packages for testing are named "fake", e.g. `k8s.io/client-go/dynamic/fake` or `FakeDiscovery{}` 😉

david-doit-intl: Fine then no worries

PR #143, opened 2021-05-07
chore: Make UPX go faster

david-doit-intl:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ pack: $(PACKED_BINS)
 
 $(PACKED_DIR)/%-$(BIN_ARCH): $(BIN_DIR)/%-$(BIN_ARCH)
 	mkdir -p $(PACKED_DIR)
-	$(UPXCMD) --brute -f -o "$@" "$<" \
+	$(UPXCMD) --lzma --best -f -o "$@" "$<" \
 	&& touch "$@"
 
 ## Create release artifacts


stepanstipl:
Can we get a more detailed analysis of the impact on size for different binaries here? The "10MB" seems to be a fit rough (from latest nightly):
```bash
$ ls -lh  kubent*
-rwxr-xr-x@ 1 stepan  staff   6.7M  5 May 03:43 kubent
-rw-r--r--@ 1 stepan  staff   6.7M  7 May 11:26 kubent-nightly-0.4.0-7-ge158834-linux-amd64.tar.gz
```

david-doit-intl:
```bash
du -sh release-artifacts/*
8.3M    kubent
8.1M    kubent-dev-linux-amd64.tar.gz
```
I just removed brute and set it fastest. I think this is reasonable comprise.

david-doit-intl:
Whilst I haven't removed UPX entirely, the builds are much faster.

stepanstipl:
What about checking previous runs? What brute does is effectively runs over all (or almost, there's `--extra-brute` too 😉 ) possible combinations, and then picks the best result. So we should be able to check which ones worked best from previous runs and use that without any impact on size. I guess this might be different for different platforms 🤔, but worth checking I think.

david-doit-intl:
Binaries are roughly 8-9 MB each

david-doit-intl:
--extra-brute way too long :robot:

stepanstipl:
hint:
```bash 
$ upx --fileinfo kubent                                                                                                                                           0
                       Ultimate Packer for eXecutables
                          Copyright (C) 1996 - 2020
UPX 3.96        Markus Oberhumer, Laszlo Molnar & John Reiser   Jan 23rd 2020

kubent [amd64-linux.elf, linux/amd64]
   7047840 bytes, compressed by UPX 13, method 14, level 10, filter 0x00/0x00
```

david-doit-intl:
Found it:

Level: 10
Method: 14

--lzma <-

stepanstipl (approved):
Looks good 👍

PR #144, opened 2021-05-07
feat: Build arm for linux and darwin

david-doit-intl:
ARM for arm linux + mac machines, closes #144


stepanstipl (commented):
Looks good, a small change in the step naming.

Could we also move the `closes #20` part of the commit message away from summary line to body/footer as per [Commit messages](https://github.com/doitintl/kube-no-trouble#commit-messages).

stepanstipl (approved):
Looks good, I've adjusted the required builds as they changed names. 👍

PR #156, opened 2021-06-10
feat: Add coverage for deprecated Storage resources - `storage.k8s.io/v1beta1` API group

stepanstipl:
This PR adds coverage for deprecated Storage resources - `storage.k8s.io/v1beta1` API group.


david-doit-intl (approved):
LGTM!

PR #157, opened 2021-06-10
fix: Fix for wrong file path for downloaded archive in integration test

david-doit-intl:

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -91,7 +91,7 @@ jobs:
         cluster_name: kubent-test-cluster
     - name: run integration test
       run: |
-        tar xvzf release-artifacts/kubent-${{ github.event.number }}merge-linux-amd64.tar.gz
+        tar xvzf release-artifacts/kubent-*-linux-amd64.tar.gz
         kubectl version --short
         kubectl cluster-info --context kind-kubent-test-cluster
         ./kubent


.github/workflows/main.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
         cluster_name: kubent-test-cluster
     - name: run integration test
       run: |
-        tar xvzf release-artifacts/kubent-${{ github.event.number }}merge-linux-amd64.tar.gz
+        tar xvzf release-artifacts/kubent-${{ github.event.number }}*-linux-amd64.tar.gz
```
stepanstipl, line 94:
```suggestion
        tar xvzf release-artifacts/kubent-*-linux-amd64.tar.gz
```

Based on previous runs the names needed to be covered are:
- for PRs: `kubent-152merge-linux-amd64.tar.gz`
- for nightly: `kubent-nightly-0.3.2-100-g1b3913b-linux-amd64.tar.gz`
- for master: `kubent-master-linux-amd64.tar.gz`
- for tags: `kubent-0.4.0-linux-amd64.tar.gz`

It seems this would work, as the `${{ github.event.number }}` will be expanded to `` (empty) in case it does not exists (like when triggered in master), but it feels a bit redundant and ugly to rely on this behaviour. I think it's not needed.

david-doit-intl: done!

stepanstipl (approved):
👍 LGTM

Issue #135, opened 2021-04-23
Review and add rules for Deprecated API Migration Guide


stepanstipl:
K8s docs now have handy page [Deprecated API Migration Guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/).

Go over it and make sure we support those resources or add them otherwise.

stepanstipl:
The last of the PRs covering resources in the Migration guide has been merged.

PR #159, opened 2021-06-10
feat: Cover deprecated `scheduling.k8s.io/v1beta1` - PriorityClass resource

stepanstipl:
Cover deprecated `scheduling.k8s.io/v1beta1` - PriorityClass resource

As per https://kubernetes.io/docs/reference/using-api/deprecation-guide/ add PriorityClass resource.

Review after #156, it will be easier :)
Part of #135


david-doit-intl (approved):
:+1:

PR #161, opened 2021-06-11
ref: Continue if individual manifest fails

### LINKED ISSUE #21: Continue processing other manifests, even if one fails
Even if one or more manifests fail to be parsed, we should still continue to process others. This is improvement after adding better error reporting in #20 to address #13 

This will require some ref. around handling errors, I imagine the collector should return multiple "non-fatal" errors in this case, and these should be logged to user.

david-doit-intl:
Closes #21 

### DIFF
--- a/pkg/collector/file.go
+++ b/pkg/collector/file.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 
 	"github.com/ghodss/yaml"
+	"github.com/rs/zerolog/log"
 	"helm.sh/helm/v3/pkg/releaseutil"
 )
 
@@ -73,7 +74,8 @@ func (c *FileCollector) Get() ([]map[string]interface{}, error) {
 				var manifest map[string]interface{}
 				err := yaml.Unmarshal([]byte(manifests[k]), &manifest)
 				if err != nil {
-					return nil, fmt.Errorf("failed to parse file %s: %v", f, err)
+					log.Warn().Msgf("failed to parse file %s: %v", f, err)
+					continue
 				}
 
 				results = append(results, manifest)
--- a/pkg/collector/file_test.go
+++ b/pkg/collector/file_test.go
@@ -62,7 +62,6 @@ func TestFileCollectorGet(t *testing.T) {
 
 func TestFileCollectorGetUnknown(t *testing.T) {
 	input := []string{"../../fixtures/meow.txt"}
-	expected := "failed to parse"
 
 	c, err := NewFileCollector(
 		&FileOpts{Filenames: input},
@@ -72,12 +71,10 @@ func TestFileCollectorGetUnknown(t *testing.T) {
 		t.Errorf("Expected to succeed for %s, failed: %s", input, err)
 	}
 
-	_, err = c.Get()
+	result, _ := c.Get()
 
-	if err == nil {
-		t.Errorf("Expected error with unknown file type")
-	} else if !strings.Contains(err.Error(), expected) {
-		t.Errorf("Expected error message with %s, got %s", expected, err.Error())
+	if len(result) != 0 {
+		t.Errorf("Expected empty result instead got %s", result)
 	}
 }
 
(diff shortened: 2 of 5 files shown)


david-doit-intl:
@stepanstipl  I thought about change the function signature for Get() =>

from:
```go
type Collector interface {
	Get() ([]map[string]interface{}, error)
	Name() string
}
```
to
```go 
type Collector interface {
	Get() []map[string]interface{}
	Name() string
}
```

however there maybe some errors we wish to return which are not related to manifest parsing.

pkg/collector/file.go
```diff
@@ -48,7 +48,8 @@ func (c *FileCollector) Get() ([]map[string]interface{}, error) {
 			input, err = ioutil.ReadFile(f)
 		}
 		if err != nil {
-			return nil, fmt.Errorf("failed to read file %s: %v", f, err)
+			log.Info().Msgf("failed to read file %s: %v", f, err)
+			continue
```
stepanstipl, line 52: If a file, that was explicitly passed via flag by the user, is not found, we should fail. I wouldn't change this - this is equivalent to failing to connect to the cluster in other collectors (or similar).

pkg/collector/file.go
```diff
@@ -73,7 +74,8 @@ func (c *FileCollector) Get() ([]map[string]interface{}, error) {
 				var manifest map[string]interface{}
 				err := yaml.Unmarshal([]byte(manifests[k]), &manifest)
 				if err != nil {
-					return nil, fmt.Errorf("failed to parse file %s: %v", f, err)
+					log.Info().Msgf("failed to parse file %s: %v", f, err)
```
stepanstipl, line 77: Maybe `Warning` is in place?

pkg/collector/helm2.go
```diff
@@ -78,8 +78,8 @@ func (c *HelmV2Collector) Get() ([]map[string]interface{}, error) {
 
 			err := yaml.Unmarshal([]byte(m), &manifest)
 			if err != nil {
-				err := fmt.Errorf("failed to parse release %s/%s: %v", r.Namespace, r.Name, err)
-				return nil, err
+				log.Info().Msgf("failed to parse release %s/%s: %v", r.Namespace, r.Name, err)
```
stepanstipl, line 81: Warning?

pkg/collector/helm3.go
```diff
@@ -77,8 +77,8 @@ func (c *HelmV3Collector) Get() ([]map[string]interface{}, error) {
 
 			err := yaml.Unmarshal([]byte(m), &manifest)
 			if err != nil {
-				err := fmt.Errorf("failed to parse release %s/%s: %v", r.Namespace, r.Name, err)
-				return nil, err
+				log.Info().Msgf("failed to parse release %s/%s: %v", r.Namespace, r.Name, err)
```
stepanstipl, line 80: Warning?

stepanstipl (commented):
Thanks 👍 .

Some minor tweaks - in general, what about using `Warn` level messages, to be consistent with existing code + it seems to be in place.

However, the PR is missing the main reason for the original #21 - when the `last-applied-configuration` can't be parsed as valid json, please add (in cluster collector - https://github.com/doitintl/kube-no-trouble/blob/055df4e213a394ad107021724714c5130759b284/pkg/collector/cluster.go#L125-L126).

david-doit-intl:
@stepanstipl updated

pkg/collector/file_test.go
```diff
@@ -93,12 +89,10 @@ func TestFileCollectorGetNonExistent(t *testing.T) {
 		t.Errorf("Expected to succeed for %s, failed: %s", input, err)
 	}
 
-	_, err = c.Get()
+	result, _ := c.Get()
```
stepanstipl, line 92: I think this test should be reverted as well.

stepanstipl (commented):
Looks good 👍 , I think you forgot to revert one test .. so just sort that one.

stepanstipl:
LGTM 👍

stepanstipl (approved):
LGTM 👍

Issue #113, opened 2021-03-04
Rules tests


ghost:
Write test to see how good rules.

stepanstipl:
I think test committed in #111 can be a starting point. Could be refactored into a matrix type of test, but this is how I would see these tests working - a list of tests manifest, paired with expected findings, should include negative test cases too.

stepanstipl:
Done.

PR #164, opened 2021-06-11
fix: Remove `LeaseList` and add `Lease` to cluster collector, add test for `Lease`

stepanstipl:
This PR is about minor fixes and adding test for `coordination.k8s.io/v1beta1` API group.

Changes:
- fix: Add Leases to be pulled by cluster collector
- fix: Remove LeaseList from rules
- test: Add test for Lease

#135, #113


david-doit-intl (approved):
LGTM

PR #182, opened 2021-06-29
feat: Filter non-relevant results

### LINKED ISSUE #130: Detect K8s version and suggest only relevant issues

stepanstipl:
This PR implements filtering of non-relevant versions, i.e. versions that the user can't upgrade to as they're not available in the used K8S version.

Closes #130

### DIFF
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -76,14 +76,21 @@ func initCollectors(config *config.Config) []collector.Collector {
 	return collectors
 }
 
-func getServerVersion(collectors []collector.Collector) (string, error) {
-	for _, c := range collectors {
-		versionCol, ok := c.(collector.VersionCollector)
-		if ok {
-			return versionCol.GetServerVersion()
+func getServerVersion(cv *config.Version, collectors []collector.Collector) error {
+	if cv.Version == nil {
+		for _, c := range collectors {
+			if versionCol, ok := c.(collector.VersionCollector); ok {
+				goversion, err := versionCol.GetServerVersion()
+				if err != nil {
+					return fmt.Errorf("failed to detect k8s version: %w", err)
+				}
+
+				cv.SetFromVersion(goversion)
+				return nil
+			}
 		}
 	}
-	return "", nil
+	return nil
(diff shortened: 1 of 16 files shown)


cmd/kubent/main.go
```diff
@@ -76,14 +76,22 @@ func initCollectors(config *config.Config) []collector.Collector {
...
-func getServerVersion(collectors []collector.Collector) (string, error) {
-	for _, c := range collectors {
-		versionCol, ok := c.(collector.VersionCollector)
-		if ok {
-			return versionCol.GetServerVersion()
+func getServerVersion(cv *config.Version, collectors []collector.Collector) error {
+	if cv.Version == nil {
+		for _, c := range collectors {
+			versionCol, ok := c.(collector.VersionCollector)
+			if ok {
```
david-doit-intl, line 83:
```
if versionCol, ok := c.(collector.VersionCollector); ok {
```

stepanstipl: done.

cmd/kubent/main_test.go
```diff
@@ -146,13 +147,15 @@ func TestMainExitCodes(t *testing.T) {
 func TestGetServerVersionNone(t *testing.T) {
 	collectors := []collector.Collector{}
 
-	version, err := getServerVersion(collectors)
+	version := config.NewVersion()
+
+	err := getServerVersion(version, collectors)
 	if err != nil {
```
david-doit-intl, line 153: can do a 
```go
if err := = getServerVersion(version, collectors); err != nil {
```

david-doit-intl (approved):
LGTM, I like these changes!

cmd/kubent/main_test.go
stepanstipl: done.

PR #196, opened 2021-07-23
Fix: Helm v3 collector name and refactoring to be test-friendly

stepanstipl:
The main reason for this PR is to fix Helm v3 collector name:

```diff
pkg/collector/helm3.go
- commonCollector: newCommonCollector("Helm v2"),
+ commonCollector: newCommonCollector("Helm v3"),
```

And I've also added a test to cover this, and that needed some small refactoring too :).

### DIFF
--- a/pkg/collector/helm2_test.go
+++ b/pkg/collector/helm2_test.go
@@ -1,0 +1,28 @@
+package collector
+
+import (
+	"testing"
+
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestNewHelmV2Collector(t *testing.T) {
+	expectedName := "Helm v2"
+
+	clientSet := fake.NewSimpleClientset()
+	col, err := NewHelmV2Collector(&HelmV2Opts{
+		DiscoveryClient: clientSet.Discovery(),
+		CoreClient:      clientSet.CoreV1(),
+	})
+	clientSet.CoreV1()
+
+	if err != nil {
+		t.Fatalf("Failed to create collector from fake discovery client")
+	}
+	if col == nil {
+		t.Fatalf("Should return collector, got nil instead")
+	}
+	if col.Name() != expectedName {
+		t.Fatalf("Expected collector name: %s, instead got: %s", expectedName, col.Name())
+	}
+}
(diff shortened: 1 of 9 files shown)


pkg/collector/helm3.go
```diff
@@ -30,12 +31,13 @@ func NewHelmV3Collector(opts *HelmV3Opts) (*HelmV3Collector, error) {
 	}
 
 	collector := &HelmV3Collector{
-		commonCollector: newCommonCollector("Helm v2"),
+		commonCollector: newCommonCollector("Helm v3"),
```
david-doit-intl, line 34: Should the string be a const so you can refer to it in the test?

pkg/collector/helm2_test.go
```diff
@@ -0,0 +1,28 @@
+package collector
+
+import (
+	"testing"
+
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestNewHelmV2Collector(t *testing.T) {
+	expectedName := "Helm v2"
```
david-doit-intl, line 10: I'd consider making this a const in pkg/collector/helm2.go, to make the tests less brittle

pkg/collector/helm3.go
stepanstipl: :) Yeah, I was thinking about it... it would be nicer in the code, but then I'm not sure from the test perspective. As in the test then wouldn't catch if it gets changed to something undesirable. So 🤷‍♂️ ...

Happy to change if you think it's better, I really don't have a strong opinion here.

david-doit-intl (changes requested):
LGTM, just a couple very slight changes

pkg/collector/helm3.go
david-doit-intl: Yeah I like the const. But yeah not a very strong opinion either way.

stepanstipl: Yeah, changed it, consts look nicer 👍

pkg/collector/helm2_test.go
stepanstipl: done

stepanstipl:
_Just for the record, I'm going to intentionally ignore SonarCloud here, and its warning about code duplication in `pkg/collector/helm2.go` and `pkg/collector/helm3.go` While there's some code dupe (and we probably should look at in future), this PR does not make it worse and refactoring that is out of scope for now._

david-doit-intl (approved):

PR #198, opened 2021-07-28
fix: Fix nil namespace helm

### LINKED ISSUE #124: panic: interface conversion: interface {} is nil, not string
Running into the below issue while running kubent, can someone help?

```

goroutine 1 [running]:
github.com/doitintl/kube-no-trouble/pkg/judge.(*RegoJudge).Eval(0xc00015fc10, 0xc000273500, 0x29c, 0x2a0, 0x29c, 0x2a0, 0x0, 0x1f4, 0xc0002f4390)
	/__w/kube-no-trouble/kube-no-trouble/pkg/judge/rego.go:76 +0x7c2
main.main()
	/__w/kube-no-trouble/kube-no-trouble/cmd/kubent/main.go:97 +0x3be
```

Go version - go1.15.5 darwin/amd64

kubent version - 0.3.2

stepanstipl:
This PR should fix #124, after some tests the only way the OPA expressions will return nil namespace is if input has namespace nil / `null` in json. It seems it might be the case with Helm and interface assertions.

Used this opportunity to extract out some dupe code, and cover it with tests.

The `TestFixNamespace/nil` test was first added and failing before the fix.

### DIFF
--- a/pkg/collector/helm.go
+++ b/pkg/collector/helm.go
@@ -33,7 +33,7 @@ func fixNamespace(manifest *map[string]interface{}, defaultNamespace string) {
 	if meta, ok := (*manifest)["metadata"]; ok {
 		switch v := meta.(type) {
 		case map[string]interface{}:
-			if _, ok := v["namespace"]; !ok {
+			if val, ok := v["namespace"]; !ok || val == nil {
 				v["namespace"] = defaultNamespace
 			}
 		}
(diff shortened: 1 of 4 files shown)


stepanstipl:
_Again, the Sonar dupe code check sucks here. The PR is actually reducing code duplication, and the flagged part of code can't be easily deduplicated, as it's dealing with another typers - `helm.sh/helm` vs `helm.sh/helm/v3` `Release`._

david-doit-intl (approved):
:pray:

PR #203, opened 2021-08-09
docs: Add logos

stepanstipl:
Add logos for the project.


stepanstipl:
@david-doit-intl I'm going to abuse my admin rights here and merge this, as I know you're away for a while, and since this is just docs I think it's ok, but if you have any concerns let me know & happy to get back to this _(merging without review should not be common practice here)._

PR #213, opened 2021-09-03
fix: Output 'since' correctly in JSON and text formats

### LINKED ISSUE #207: Output for "Since" is broken
There are several issues with output of `Since` field in findings:

1) JSON output is broken:

```
$./bin/kubent-darwin-amd64 -o json
...
[
        {
                "Name": "master-monitoring-role-binding",
                "Namespace": "\u003cundefined\u003e",
                "Kind": "ClusterRoleBinding",
                "ApiVersion": "rbac.authorization.k8s.io/v1beta1",
                "RuleSet": "Deprecated APIs removed in 1.22",
                "ReplaceWith": "rbac.authorization.k8s.io/v1",
                "Since": {}
        },
```

2) JSON output for additional kinds (when Since is nil) is null:

```$./bin/kubent-darwin-amd64 -a AllowlistedWorkload.v1.auto.gke.io -o json
...
        {
                "Name": "cc",
                "Namespace": "\u003cundefined\u003e",
                "Kind": "AllowlistedWorkload",
                "ApiVersion": "auto.gke.io/v1",
                "RuleSet": "Additional resources (custom)",
                "ReplaceWith": "\u003cna\u003e",
                "Since": null
        },
```

3) Text output when Since is nil is `<nil>`:

```
$./bin/kubent-darwin-amd64 -a AllowlistedWorkload.v1.auto.gke.io 
... 
>>> Additional resources (custom) <<<
KIND                  NAMESPACE     NAME             API_VERSION      REPLACE_WITH (SINCE)
AllowlistedWorkload   <undefined>   cc               auto.gke.io/v1   <na> (<nil>+)
AllowlistedWorkload   <undefined>   datadog-agents   auto.gke.io/v1   <na> (<nil>+)
```

1) Should be fixed, 2) is to decide - either omit field or an empty string, 3) to be fixed.

TyeMcQueen:
Fixes #207

### DIFF
--- a/pkg/judge/judge.go
+++ b/pkg/judge/judge.go
@@ -11,7 +11,8 @@ type Result struct {
 	ApiVersion  string
 	RuleSet     string
 	ReplaceWith string
-	Since       *goversion.Version
+	Since       *goversion.Version `json:"-"`
+	SinceStr    string             `json:"Since"`
 }
 
 type Judge interface {
--- a/pkg/judge/rego.go
+++ b/pkg/judge/rego.go
@@ -53,9 +53,12 @@ func (j *RegoJudge) Eval(input []map[string]interface{}) ([]Result, error) {
 				m := i.(map[string]interface{})
 				log.Trace().Msgf("parsing +%v", m)
 
+				sinceStr := ""
 				since, err := goversion.NewVersion(m["Since"].(string))
 				if err != nil {
 					log.Debug().Msgf("Failed to parse version: %s", err)
+				} else {
+					sinceStr = since.String()
 				}
 
 				results = append(results, Result{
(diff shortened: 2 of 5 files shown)


stepanstipl:
Hi @TyeMcQueen, first thanks a lot for contributing! 👍 

I have reviewed the code, and I think the duplication of the information is a bit unfortunate:
```go
	Since       *goversion.Version `json:"-"`
	SinceStr    string             `json:"Since"`
```

This would bring issues of keeping the info in sync (I know it's currently written only at one place :)).

I would suggest using embedding and creating our own type would be the way forward, i.e. smth. like:
```go
type Version struct {
    *goversion.Version
}
```

This would allow us to directly use all the existing methods, i.e. keep the code pretty much unchanged, as well as implement a new method to satisfy go JSON encoder: 
```go
func (v Version) MarshalText() ([]byte, error) {
...
}
```

What do you think?

TyeMcQueen:
Good point.  I will try to get to that soon.

david-doit-intl:
this should have been fixed by #225 I am going to close this.

PR #219, opened 2021-09-24
feat: Golang 1.17

david-doit-intl:
Bumps golang from 1.16.5-alpine3.12 to 1.17beta1-alpine3.12.

---
updated-dependencies:
- dependency-name: golang
  dependency-type: direct:production
...

Signed-off-by: dependabot[bot] <support@github.com>


stepanstipl:
Looks good, but can we pls. change the commit message? I think `dep:` in this case.

stepanstipl (changes requested):
Looks good, but can we pls. change the commit message? I think `dep:` in this case.

david-doit-intl:
> Looks good, but can we pls. change the commit message? I think `dep:` in this case.

done

PR #225, opened 2021-10-20
Fix since output format

stepanstipl:
This PR fixes `Since` format in output (#207) by wrapping goversion in our own type (which already existed for config purposes) and implementing custom marshal/unmarshal fns. Also adds tests for existing `Version` code.

Changes included:

- Wrap goversion with internal Version in judge pkg (already existed in config)
- Add tests for original `Version`
- Refactors `getServerVersion` to pure function - this is to be able to use `nil` semantics to represent unset value
- Fixes `Since` output format by implementing `(Un)MarshalText`


stepanstipl:
Yeah, agreed, there's a bit of refactor - it's separated on commits level (`fix:` vs `ref:`).

Also, I plan to ignore the remaining SonarCloud issues against duplication (unless objections) - it's in tests:
- Define a constant instead of duplicating this literal "expected: %q, got: %q" 7 times.
- Define a constant instead of duplicating this literal "expected to succeed, failed instead: %v" 3 times.

stepanstipl:
🤷‍♂️ no clue, at first look it all seems to be UI based. I would like to move most of this in the make - i.e. add some go configurable go meta-linter tool to run as part of `make test`.

Issue #179, opened 2021-06-18
Virus scanning as part of release process


stepanstipl:
While we're conscious about SDLC process security, and therefore highly unlikely we might consider adding a virus scan as an additional step to our release process - related to #178.

PR #226, opened 2021-10-22
build: Added virus scan

ghost:
Closes https://github.com/doitintl/kube-no-trouble/issues/179

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -59,6 +59,35 @@ jobs:
         with:
           name: release-artifacts-${{ matrix.os }}-${{ matrix.arch }}
           path: release-artifacts
+  virus-scan:
+    name: virus-scan
+    needs:
+      [build]
+    runs-on: ubuntu-latest
+    container: debian
+    strategy:
+      matrix:
+        os: [linux, darwin, windows]
+        arch: [arm64, amd64]
+        exclude:
+          - os: windows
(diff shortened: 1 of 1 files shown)


david-doit-intl:
@stepanstipl this is ready, but let me know if its even needed

stepanstipl:
Thanks, I think it looks good in general, but I would prefer to move logic from GH pipelines to Makefile, in order to be able to run these steps easily and avoid introducing another env (debian container with a bunch of installed things).

My suggestions (assuming ClamAV can run fine on alpine):
- Remove the `virus-scan` job
- Move the required software deps to `scripts/alpine-setup.sh`, probably incl. the `freshclam`
- Create a new Makefile target `test-virus` (or `scan-virus`? but not sure we need new prefix)
- That target scans all files (or better the relevant binary only) in BIN_DIR
  - will need to start & stop ClamAV (if not running)
  - but no need to unpack files here (actually the Make target should have proper deps. on the built binaries)
- Add this as a new step to the existing `build` job

I think this should save some work, code, and make it runnable locally and probably faster in the CI.


And checking if ClamAV support scanning archives, which it seems [it does](https://linux.die.net/man/1/clamscan) (`--scan-archive` flag, should be on by default), maybe scan the whole archive rather than binaries (probably doesn't matter, but in case we would decide to add anything in future and the archive is the end product we distribute).

PR #237, opened 2021-11-18
Add support for in-cluster auth

stepanstipl:
This PR mainly adds support for in-cluster authentication and some minor refactoring on tests.

This func. would deserve integration test, created #236.

### DIFF
--- a/pkg/collector/kube.go
+++ b/pkg/collector/kube.go
@@ -26,7 +26,7 @@ func newKubeCollector(kubeconfig string, kubecontext string, discoveryClient dis
 			return nil, fmt.Errorf("failed to assemble client config: %w", err)
 		}
 
-		if discoveryClient, err = discovery.NewDiscoveryClientForConfig(col.restConfig); err != nil {
+		if col.discoveryClient, err = discovery.NewDiscoveryClientForConfig(col.restConfig); err != nil {
 			return nil, fmt.Errorf("failed to create client: %w", err)
 		}
 	}
--- a/pkg/collector/kube.go
+++ b/pkg/collector/kube.go
@@ -21,30 +21,48 @@ func newKubeCollector(kubeconfig string, kubecontext string, discoveryClient dis
 	if discoveryClient != nil {
 		col.discoveryClient = discoveryClient
 	} else {
-		pathOptions := clientcmd.NewDefaultPathOptions()
-		if kubeconfig != "" {
-			pathOptions.GlobalFile = kubeconfig
+		var err error
+		if col.restConfig, err = newClientRestConfig(kubeconfig, kubecontext, rest.InClusterConfig); err != nil {
+			return nil, fmt.Errorf("failed to assemble client config: %w", err)
 		}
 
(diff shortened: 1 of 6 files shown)


pkg/collector/kube.go
```diff
@@ -21,30 +21,48 @@ func newKubeCollector(kubeconfig string, kubecontext string, discoveryClient dis
 	if discoveryClient != nil {
 		col.discoveryClient = discoveryClient
 	} else {
-		pathOptions := clientcmd.NewDefaultPathOptions()
-		if kubeconfig != "" {
-			pathOptions.GlobalFile = kubeconfig
+		var err error
```
david-doit-intl, line 24: is this var needed?

david-doit-intl (approved):

pkg/collector/kube.go
stepanstipl: Kind of yes, as it is, if I don't declare the variable, the `if` below would fail on undefined var. The issue is the choice of operator `:=` vs `=`, and so both of the variables on the left side have to be either undefined or exist. I guess I can refactor both to:
```go
if restConfig, err := newClientRestConfig(kubeconfig, kubecontext, rest.InClusterConfig); err != nil {
    return nil, fmt.Errorf("failed to assemble client config: %w", err)
}
col.restConfig = restconfig
```

In the end it will result in extra line, as it would have to be done for both `if`s, but perhaps it's nicer? WDYT?

stepanstipl:
Btw. test this works:
```
kubent-linux-amd64
2:23PM INF >>> Kube No Trouble `kubent` <<<
2:23PM INF version dev (git sha dev)
2:23PM INF Initializing collectors and retrieving data
2:23PM INF Target K8s version is 1.21.1
2:23PM INF Retrieved 0 resources from collector name=Cluster
2:23PM ERR Failed to retrieve data from collector error="secrets is forbidden: User \"system:serviceaccount:default:default\" cannot list resource \"secrets\" in API group \"\" at the cluster scope" name="Helm v2"
2:23PM ERR Failed to retrieve data from collector error="list: failed to list: secrets is forbidden: User \"system:serviceaccount:default:default\" cannot list resource \"secrets\" in API group \"\" at the cluster scope" name="Helm v3"
2:23PM INF Loaded ruleset name=custom.rego.tmpl
2:23PM INF Loaded ruleset name=deprecated-1-16.rego
2:23PM INF Loaded ruleset name=deprecated-1-22.rego
2:23PM INF Loaded ruleset name=deprecated-1-25.rego
```
(the errors are expected due to limited permissions of default account that got picked up)

And once more planning to ignore Sonar :).

pkg/collector/kube.go
david-doit-intl: I prefer not having a var, and using:

```
:=
```

stepanstipl: For the record settled on keeping as is - my example above wouldn't work, and with `:=` one would have to drop the short form, as otherwise the `restConfig` would be in local if scope only.

PR #268, opened 2022-01-26
feat: Version flag

### LINKED ISSUE #259: Add a version command
Hi,

would it be possible to add a `kubent --version` or `kubent version` command to check the current version?

I'd like to use that command to verify that the installation was correctly. Currently, when running `kubent version` it'll start `kubent` without any error or waning :D

dark0dave:
Closes #259

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,6 +24,7 @@ type Config struct {
 	LogLevel        ZeroLogLevel
 	Output          string
 	TargetVersion   *judge.Version
+	KubentVersion   bool
 }
 
 func NewFromFlags() (*Config, error) {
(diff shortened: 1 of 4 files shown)


dark0dave:
needs #269

pkg/config/config.go
```diff
@@ -36,6 +37,7 @@ func NewFromFlags() (*Config, error) {
 	flag.BoolVarP(&config.Cluster, "cluster", "c", true, "enable Cluster collector")
 	flag.StringVarP(&config.Context, "context", "x", "", "kubeconfig context")
 	flag.BoolVarP(&config.ExitError, "exit-error", "e", false, "exit with non-zero code when issues are found")
+	flag.BoolVarP(&config.KubentVersion, "version", "v", false, "fetch version information of kubent binary")
```
stepanstipl, line 40: This doesn't match the output added to the README. I would prefer that formulation:
```
prints the version of kubent and exits
```

stepanstipl (commented):
Looks good, just align the above. 👍

pkg/config/config.go
dark0dave: Thats actually what I had first. Done!

stepanstipl (approved):
Cool, thx. 👍 🎉

PR #269, opened 2022-01-26
chore: Pre-commit fixed versions and remove broken make linter

dark0dave:


stepanstipl (approved):
👍 Looks like the `checkmake` wasn't doing much anyway, so I think it's ok to remove.

PR #302, opened 2022-06-20
feat: kubernetes CronJob with prometheus output + grafana dashboard

milanholubstratox:


docs/manifests/kubent-cronjob.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
stepanstipl, line 1: Can we add `Namespace` resource for `kubent-system` too? I realise that you expect the Prometheus push gateway anyway, but I think it would be nice to have this working out of the box (I believe `apply` would fail if the namespace is not present currently).

docs/manifests/kubent-cronjob.yaml
```diff
@@ -0,0 +1,64 @@
...
+          containers:
+          - name: jq-curl
+            image: apteno/alpine-jq:2022-06-12
+            command: 
+              - /bin/sh
+              - -ec
+              - |
+                cat /cache/scan.json |
+                jq -r '.[]|"kubent{"+([(to_entries[]|[((.key|ascii_downcase), .value)]|join("=\""))]|join("\", ")+"\"} 1")' |
+                curl -X PUT --data-binary @- http://prometheus-operator-prometheus-pushgateway.kubent-system.svc.cluster.local:9091/metrics/job/kubent
```
stepanstipl, line 49: This is just a nitpick, but maybe it would be nice to use smth. like `$(PROMETHEUS_GATEWAY_HOST)` and `$(PROMETHEUS_GATEWAY_PORT)` and corresponding env. variables to make the target destination more obvious in case someone wants to change it.

stepanstipl (commented):
Hi @milanholubstratox,

many thanks for this, this is beautiful. 👏  Very similar to what I had in mind, and definitely happy to add this as an example, I think it will be valuable for others.

I've commented on some small improvements, pls. have a look. I also have some other bits I would like to see in future in mind, but these require work outside the scope of this, so I've created corresponding issues and we can address these in future.
- Use the official image - depends on #303 
- I would like to drop the shell from the docker image, but I understand why it's currently needed - #304
- Maybe add some integration test to make sure this works as expected - could be smth. simple like - deploy the Prom push gateway, and the manifests, and check we have kubent metrics in prometheus (this might be a bit over, but in general I believe it's good idea to have things covered with tests to avoid them getting stale).

So please have a look at the minor 2 comments in the code and let me know what you think, and otherwise, I'm happy to merge it pretty much as it is now! Thanks again 👍

Dockerfile
```diff
@@ -14,12 +14,10 @@ COPY . .
 
 RUN make all
 
-FROM scratch
+FROM alpine:3.16.0
```
dark0dave, line 17: why not scratch?

stepanstipl:
> I'd like to keep it running in scratch

@dark0dave see my notes above, in particular:

> I would like to drop the shell from the docker image, but I understand why it's currently needed - https://github.com/doitintl/kube-no-trouble/issues/304

... I believe currently the shell (therefore alpine) is needed to redirect output to a file. I agree to keep scratch, and so #304, but I think that's something that can be done before we start publishing the official docker image.

docs/manifests/kubent-cronjob.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  namespace: kubent-system
+  name: kubent-reader
+rules:
+- apiGroups: [""]
+  resources: ["configmaps", "secrets"]
```
kkopachev, line 8: Just wanted to say that this is only enough for helm. Cluster collector silently ignores access denied and reports 0 resources discovered. 

```suggestion
- apiGroups: ["*"]
  resources: ["*"]
```

Issue #283, opened 2022-03-18
Running install-kubent fails on linux/arm64 


pedantic79:
Since kubent has a binary available for linux/arm64. I tried to install it via the `install-kubent `script.
However because `uname -m` on Linux returns `aarch64` it fails. On MacOS, `uname -m` returns `arm64`.

```
# uname -s -m
Linux aarch64
# sh -c "$(curl -sSL https://git.io/install-kubent)"
>>> kubent installation script <<<
> Detecting latest version
> Downloading version 0.5.1
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     9  100     9    0     0     35      0 --:--:-- --:--:-- --:--:--    35
tar: invalid magic
tar: short read
```

There is a workaround using the `TARGET_ARCH` env, but it would be nice if the script supported this directly.

```
# uname -s -m
Linux aarch64
# TARGET_ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/aarch64/arm64/')" sh -c "$(curl -sSL https://git.io/install-kubent)"
>>> kubent installation script <<<
> Detecting latest version
> Downloading version 0.5.1
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   670  100   670    0     0   2401      0 --:--:-- --:--:-- --:--:--  2401
100 8808k  100 8808k    0     0  5532k      0  0:00:01  0:00:01 --:--:-- 13.0M
> Done. kubent was installed to /usr/local/bin/.
```

PR #284, opened 2022-03-18
fix: Arch detection for aarch64

Staphylo:
Use the same sed trick for x86_64 -> amd64

Fixes #283

PR #305, opened 2022-06-22
fix: Arch detection dectection for aarch64

dark0dave:
Use the same sed trick for x86_64 -> amd64

Fixes https://github.com/doitintl/kube-no-trouble/issues/283

Using: https://github.com/doitintl/kube-no-trouble/pull/284


stepanstipl (approved):
Looks good to me (haven't actually tested, so I trust you here 😆 )!

Issue #306, opened 2022-06-22
Clean up CI/CD


dark0dave:
Its been awhile since we looked at the CI/CD stuff it needs some love

dark0dave:
fatal: unsafe repository is happening alot

dark0dave:
go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.

dark0dave:
some deprecations going on

PR #308, opened 2022-06-22
ci: update checkout action to lastest

dark0dave:
Part of https://github.com/doitintl/kube-no-trouble/issues/306

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -10,7 +10,9 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3.0.2
+      with:
+        set-safe-directory: true
     - uses: actions/setup-python@v2
     - uses: pre-commit/action@v2.0.0
   test:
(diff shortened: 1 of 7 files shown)


dark0dave:
what a mess...

dark0dave:
> @dark0dave would this allow us to remove the changes in #305? In particular the
> 
> ```
> git config --global --add safe.directory "$GITHUB_WORKSPACE" || true
> ```
> 
> is causing the following issue when run locally (used to work):
> 
> ```
> $ gmake test
> test -z "$(opa fmt -l pkg/rules/rego/*)"
> test -z "$(go fmt ./...)"
> ./scripts/git-check-commits.sh
> ./scripts/git-check-commits.sh: line 10: GITHUB_WORKSPACE: unbound variable
> gmake: *** [Makefile:96: test-git] Error 1
> ```
> 
> ... so I'm wondering if this would allow us to drop that change?

So I tried removing it and because we clone it still there are still some issues.

dark0dave:
@stepanstipl fixed :)

stepanstipl (approved):
LGTM

Issue #304, opened 2022-06-22
Add config option to allow output to specific file, e.g. `--output-file`


stepanstipl:
This will allow for better integration in situations like #302, without requiring a shell in the container image.

PR #312, opened 2022-06-28
ci: Build and publish docker image

### LINKED ISSUE #303: Publish official container image
Probably use GH for hosting, GH actions for podman. Would also be nice to add corresponding make targets to do same locally.

dark0dave:
Closes #303


dark0dave:
@stepanstipl wdyt?

dark0dave:
bump @stepanstipl

stepanstipl:
Yep, so I was catching up on this. So let's first clean up the docker image, this was related to https://github.com/doitintl/kube-no-trouble/pull/302 and #304. I'm working on the the #304.

So I think the order of things is #304, resp. the PR for it -> Clean up docker image (scratch + some minor bits) -> this PR.


Otherwise first glance at the PR - I don't understand the `.github/workflows/docker.yaml`:
```
name: Publish Docker image
on:
  push:
    branches: ['release']
...
```

We don't have a branch `release`... what am I missing? I think docker image should get built as part of the normal main workflow, as it is now done, and published as part of the same workflow - perhaps part of `release-artifacts` or smth. like that, basically same as when we publish other artifacts.

dark0dave:
@stepanstipl I updated the publish action

```
on:
  release:
    types: [published]
```
this should be what we are after.

The reason for separating these is to reduce the size and complexity of the main.yaml file, whilst keeping the deployment docker image separate to the binary deployment. image is published to

stepanstipl:
Thanks, I think this is better. A couple of Q:

- Will this work for nightly pre-releases? I think we want to publish Docker image for those as well.
- While I appreciate keeping the GH workflow files simple, I think a separate trigger and workflow adds more complexity than removes here. We add use of 2 different actions and introduce a new trigger.
- Related to the above - I would like to build docker images, and include links to the built images in the generated release notes, same as links to binary artifacts are added by GH (I'm not saying this should be implemented in this PR). How would it work in this workflow? (I understand the Docker builds would trigger only after the release).
- Related to that - what if Docker build/push fails, we have a release without Docker images?


And can you share an example of the full Docker image URL and what tags would be created? 🙏

dark0dave:
- Yes it would work for nighties I can add for them too
- Well, the first trigger actually creates the release with the binaries in it, then the docker image gets published.
- We can pull them from the existing release.
- Shouldn't happen as the normal trigger for each release and main, has a docker build step in there. If we break the docker build steps we can't make a release.

dark0dave:
- [ ] Add to main workflow
- [ ] Build before create release 
- [ ] Add to release notes

dark0dave:
@stepanstipl updated

Dockerfile
```diff
@@ -9,7 +9,8 @@ COPY pkg pkg
 COPY Makefile Makefile
 RUN make all
 
-FROM alpine:3.16.0
+FROM scratch
 USER 1000
```
stepanstipl, line 13: Since we're cleaning up the Dockerfile, can we please also add a group here? (And maybe use 10k as a user, this is a nitpick, but a lesser chance of clashing with actual users on any system)

```suggestion
USER 10000:10000
```

scripts/github-changelog.sh
```diff
@@ -19,3 +19,5 @@ echo "- Generating changelog (${OUTPUT_FILE})"
 changelog-gen -changelog "${CHANGELOG_TEMPLATE}" \
 	-owner "${GH_ORG}" -repo "${GH_REPO}" \
 	"${RELEASE_SHA}" "${MASTER_SHA}" | tee "${OUTPUT_FILE}"
+
+echo "**Docker Image Perma Link**: ghcr.io/${GH_ORG}/${GH_REPO}:${RELEASE_SHA}" >> ${OUTPUT_FILE}
```
stepanstipl, line 23: Can we use the actual version tag in the link? The `${RELEASE_SHA}` would be very confusing _(should we want anything static, it should be docker image digest, but I think the release version makes most sense)._

stepanstipl: Also can we please add the docker link via the `changelog.tmpl`, instead of "echoing" to the generated file (the templating language has `env` functionality [^1], so we can pull it from env variable there).

[^1]: http://masterminds.github.io/sprig/os.html

stepanstipl (changes requested):
I think all the changes are good. Almost there ;) 🙏  Some minor bits noted in the comments. And apart from the circular dep. issue which I believe would cause the workflow to fail, only one point - are we adding the `latest` tag?

Dockerfile
dark0dave: but the user only exists in the docker image? Unless I am missing something?

dark0dave: happy to make that change

.github/workflows/main.yaml
```diff
@@ -101,10 +118,36 @@ jobs:
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: true
+          tags: |
+            kubent:${{ github.sha }},
+            kubent:${{ needs.create-release.outputs.tag_name }}
```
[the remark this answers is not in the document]
dark0dave: then it will be pushed after, we are not calculating the tag twice.

dark0dave:
> I think all the changes are good. Almost there ;) pray Some minor bits noted in the comments. And apart from the circular dep. issue which I believe would cause the workflow to fail, only one point - are we adding the `latest` tag?

I am going to skip adding the latest tag for now. Lets get this landed.

scripts/github-changelog.sh
dark0dave: yeah sorry the docs for this where very hard to find, not great.

Dockerfile
stepanstipl: 👍 and not really, while Docker supports users ns for some time, I believe these are not enabled by default [^1], i.e. user/gids are shared. I.e. it's a good security practice to avoid running as root, and other low numbers that are likely to exist on the host system.

[^1]: https://docs.docker.com/engine/security/userns-remap/

scripts/github-changelog.sh
stepanstipl, line 23: I think this is now duplicate and should be removed...

scripts/changelog.tmpl
```diff
@@ -66,3 +66,6 @@
 * {{. }}
 {{end -}}
 {{- end -}}
+
+**Docker Image Perma Link**:
```
stepanstipl, line 70: Can we just call this "Docker Image"? I mean it's an image name, not an actual link.

scripts/changelog.tmpl
```diff
@@ -66,3 +66,6 @@
...
+{{ expandenv "ghcr.io/$GH_ORG/$GH_REPO:$RELEASE_SHA" }}
```
stepanstipl, line 71: > Can we use the actual version tag in the link? The ${RELEASE_SHA} would be very confusing (should we want anything static, it should be docker image digest, but I think the release version makes most sense).

.github/workflows/main.yaml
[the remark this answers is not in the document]
stepanstipl: We could get the tag in this action, and then use it in the release step to make sure it's consistent. But I guess it's ok, let's see ;).

stepanstipl (changes requested):
I guess it's ok to leave the latest for later, we should create an issue in that case.

scripts/github-changelog.sh
dark0dave: my mistake removed

scripts/changelog.tmpl
dark0dave: Sure its both tho?

scripts/changelog.tmpl
dark0dave: Is that environment variable even exposed?

dark0dave: I think GITHUB_REF_NAME should do it @stepanstipl as defined here:

https://docs.github.com/en/actions/learn-github-actions/environment-variables

.github/workflows/main.yaml
```diff
@@ -129,6 +146,39 @@ jobs:
...
+      - name: Push latest
+        uses: docker/build-push-action@v3
+        if: "!contains(github.ref, 'nightly')"
```
dark0dave, line 177: @stepanstipl this should cover the latest

scripts/changelog.tmpl
stepanstipl: I would expect a link to have a full form (i.e. start with `http(s)://...`). And while the names are eventually translated to links, you can't really directly download docker images like `alpine`, or `us-docker.pkg.dev/google-samples/containers/gke/hello-app:1.0` for example.

scripts/changelog.tmpl
stepanstipl: I think so 🤞 _(I can never quite tell with GH actions until really trying it)_

.github/workflows/main.yaml
```diff
@@ -129,6 +146,39 @@ jobs:
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: true
+          tags: |
+            kubent:${{ github.sha }},
+            kubent:${{ needs.create-release.outputs.tag_name }}
```
stepanstipl, line 174: I would worry this might not work - it's like a mix of both the supported formats, see [^1]. I think it has to be either without comas or one-liner.


[^1]: https://github.com/docker/build-push-action#inputs

.github/workflows/main.yaml
```diff
@@ -129,6 +146,39 @@ jobs:
...
+      - name: Push latest
+        uses: docker/build-push-action@v3
+        if: ${{ !contains(github.ref, 'nightly') }}
```
stepanstipl, line 177: I think this should work 👍 , can we just remove the `${{`? It should not be needed for ifs [^1] and is consistent with other uses + a bit more readable.

[^1]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif:~:text=create%20a%20conditional.-,When%20you%20use%20expressions%20in%20an%20if%20conditional%2C%20you%20may%20omit%20the%20expression%20syntax%20(%24%7B%7B%20%7D%7D)%20because%20GitHub%20automatically%20evaluates%20the%20if%20conditional%20as%20an%20expression.%20For%20more%20information%2C%20see%20%22Expressions.%22,-Example%3A%20Only%20run

.github/workflows/main.yaml
```diff
@@ -6,6 +6,9 @@ on:
     tags:
     - "*"
   pull_request:
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
```
stepanstipl, line 11: This seems to be unused currently...

dark0dave:
> I guess it's ok to leave the latest for later, we should create an issue in that case.

Latest has been added, I had to do some research

.github/workflows/main.yaml
stepanstipl: Also based on the examples in readme [^1], the GH user name (or org name in our case) is missing in the tag names. Perhaps the earlier introduced `IMAGE_NAME` was intended to be used here?

[^1]: https://github.com/docker/build-push-action

.github/workflows/main.yaml
```diff
@@ -129,6 +146,39 @@ jobs:
...
+          tags: |
+            kubent:${{ github.sha }},
+            kubent:${{ needs.create-release.outputs.tag_name }}
+      - name: Push latest
+        uses: docker/build-push-action@v3
+        if: ${{ !contains(github.ref, 'nightly') }}
+        with:
+          context: .
+          push: true
+          tags: kubent:latest
```
stepanstipl, line 181: Same - perhaps `IMAGE_NAME`?

stepanstipl (changes requested):
I think it's pretty much there - just see the individual comments.

.github/workflows/main.yaml
dark0dave: its needed for the !

.github/workflows/main.yaml
```diff
@@ -129,6 +146,39 @@ jobs:
...
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: true
+          tags: |
+            ${{ IMAGE_NAME }}:${{ github.sha }},
```
stepanstipl, line 173: This still has a comma at the end (I saw it was removed, and then made its way back...). I think it's not correct, as mentioned previously, but I'm not 100% sure. If it's intentional let me know.

stepanstipl (changes requested):
And CI fails[^1]:

> [Invalid workflow file: .github/workflows/main.yaml#L172](https://github.com/doitintl/kube-no-trouble/actions/runs/3147009395/workflow)
The workflow is not valid. .github/workflows/main.yaml (Line: 172, Col: 17): Unrecognized named-value: 'IMAGE_NAME'. Located at position 1 within expression: IMAGE_NAME .github/workflows/main.yaml (Line: 181, Col: 17): Unrecognized named-value: 'IMAGE_NAME'. Located at position 1 within expression: IMAGE_NAME

[^1]: https://github.com/doitintl/kube-no-trouble/actions/runs/3147009395

.github/workflows/main.yaml
stepanstipl: OK, I see this was fixed in the last commit.

.github/workflows/main.yaml
stepanstipl: Ah, that makes sense. I think that's however a YAML, rather than GH, issue. I.e. it would need to be quoted, as `!` at the beginning denotes tag.

stepanstipl (approved):
Looks good to me 👍. Any bets if it will work on the first go? 🤣

PR #323, opened 2022-07-15
feat: Support for checking additional annotations

AndreasSko:
Currently, kubent only checks the `kubectl.kubernetes.io/last-applied-configuration` annotation for deprecated APIs, which is set by kubectl. If a different tool like [kapp](https://github.com/vmware-tanzu/carvel-kapp) is used, which might set other annotations, kubent won't detect those.

This PR adds support for checking additional annotations. It can be activated via the `--additional-annotations` flag.

**Example:**
```shell
kubent -t v1.22 --additional-annotations kapp.k14s.io/original
```

The change is backward compatible, as the `kubectl.kubernetes.io/last-applied-configuration` annotation will always be checked first. Only if that annotation is not set the additional annotations are tried.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,19 +76,24 @@ You can list all the configuration options available using `--help` switch:
 ```sh
 $./kubent -h
 Usage of ./kubent:
-  -a, --additional-kind strings   additional kinds of resources to report in Kind.version.group.com format
-  -c, --cluster                   enable Cluster collector (default true)
-  -x, --context string            kubeconfig context
-  -e, --exit-error                exit with non-zero code when issues are found
-  -f, --filename strings          manifests to check, use - for stdin
-      --helm2                     enable Helm v2 collector (default true)
-      --helm3                     enable Helm v3 collector (default true)
-  -k, --kubeconfig string         path to the kubeconfig file
-  -l, --log-level string          set log level (trace, debug, info, warn, error, fatal, panic, disabled) (default "info")
-  -o, --output string             output format - [text|json] (default "text")
-  -t, --target-version string     target K8s version in SemVer format (autodetected by default)
-  -v, --version                   prints the version of kubent and exits
+  -A, --additional-annotation strings   additional annotations that should be checked to determine the last applied config
+  -a, --additional-kind strings         additional kinds of resources to report in Kind.version.group.com format
+  -c, --cluster                         enable Cluster collector (default true)
+  -x, --context string                  kubeconfig context
+  -e, --exit-error                      exit with non-zero code when issues are found
+  -f, --filename strings                manifests to check, use - for stdin
+      --helm2                           enable Helm v2 collector (default true)
+      --helm3                           enable Helm v3 collector (default true)
+  -k, --kubeconfig string               path to the kubeconfig file
+  -l, --log-level string                set log level (trace, debug, info, warn, error, fatal, panic, disabled) (default "info")
+  -o, --output string                   output format - [text|json] (default "text")
+  -t, --target-version string           target K8s version in SemVer format (autodetected by default)
+  -v, --version                         prints the version of kubent and exits
+
 ```
+- *`--additional-annotation`*
+  Check additional annotations for the last applied configuration. This can be useful if a resource was applied
+  with a tool other than kubectl. The flag can be used multiple times.
 
 - *`-a, --additional-kind`*
   Tells `kubent` to flag additional custom resources when found in the specified version. The flag can be used multiple
--- a/pkg/collector/cluster.go
+++ b/pkg/collector/cluster.go
@@ -17,8 +17,9 @@ const CLUSTER_COLLECTOR_NAME = "Cluster"
 type ClusterCollector struct {
 	*commonCollector
 	*kubeCollector
-	clientSet           dynamic.Interface
-	additionalResources []schema.GroupVersionResource
+	clientSet             dynamic.Interface
+	additionalResources   []schema.GroupVersionResource
+	additionalAnnotations []string
 }
 
 type ClusterOpts struct {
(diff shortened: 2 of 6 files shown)


AndreasSko:
@stepanstipl would you be able to have a look at this PR if you got some time? What do you think about this idea? :slightly_smiling_face:

stepanstipl:
Hi @AndreasSko thanks, I was not aware of the kapp, but I like this, it only makes sense, and the PR looks good at first sight. 👍  So that's great 🎉 , thanks a lot for contributing.

I've included it in the current milestone, so will try to get to it and review it a bit more in detail soon - please allow me some time for that - but let's try to get it in with the next release.

pkg/collector/cluster.go
```diff
@@ -127,3 +132,14 @@ func (c *ClusterCollector) Get() ([]map[string]interface{}, error) {
 
 	return results, nil
 }
+
+func (c *ClusterCollector) getLastAppliedConfig(r unstructured.Unstructured) (string, bool) {
+	annotations := append(c.additionalAnnotations, "kubectl.kubernetes.io/last-applied-configuration")
```
stepanstipl, line 137: I think that contrary to the PR description, the additional annotation(s) will have priority over the `last-applied...` one, as we return with the first one found. I don't have a strong case for either, but as mentioned - for compatibility - I would prefer the other way around.
```suggestion
	annotations := append([]string{"kubectl.kubernetes.io/last-applied-configuration"}, c.additionalAnnotations...)
```

pkg/collector/cluster_test.go
```diff
@@ -104,7 +105,7 @@ func TestClusterCollectorGetFake(t *testing.T) {
 			discoveryClient := discoveryFake.FakeDiscovery{}
 			testOpts := ClusterOpts{ClientSet: clientset, DiscoveryClient: &discoveryClient}
 
-			collector, err := NewClusterCollector(&testOpts, []string{})
+			collector, err := NewClusterCollector(&testOpts, []string{}, []string{"kapp.k14s.io/original"})
```
stepanstipl, line 108: Let's not pass the additional `"kapp.k14s.io/original"` to every test run, but we can add it as another parameter to the `testCases` array, and only pass it with the relevant test.

pkg/collector/cluster.go
```diff
@@ -127,3 +132,14 @@ func (c *ClusterCollector) Get() ([]map[string]interface{}, error) {
 
 	return results, nil
 }
+
+func (c *ClusterCollector) getLastAppliedConfig(r unstructured.Unstructured) (string, bool) {
```
stepanstipl, line 136: Can we add tests for this new function? I guess it will require a similar approach with the fake client as [^1]. I would suggest testing that we get:
- `kubectl.kubernetes.io/last-applied-configuration` if present
- false if none present
- custom additional annotation if requested and present
- and that kubectl one takes precedence (this would cover the issue discussed below)

[^1]: https://github.com/doitintl/kube-no-trouble/blob/a339b08f2e91c142e2dd80026e1039d343033b87/pkg/collector/cluster_test.go#L104



[1]: https://github.com/doitintl/kube-no-trouble/blob/a339b08f2e91c142e2dd80026e1039d343033b87/pkg/collector/cluster_test.go#L104

stepanstipl (changes requested):
@andreassko thanks, I think in general it looks good 👍, good job, thanks 👏 .

Some minor comments on naming and missing a test for the new function. I hope it's all straight-forward, and if addressed, we can get it in the next release. Ping me if you have any questions. _(Also please rebase on the master, since you'll be already making changes 🙏 )_

pkg/collector/cluster.go
stepanstipl: Thinking about this a bit more, resp. about the difficult testing. One though - what if we rewrite the `getLastAppliedConfig` as:
```go
func (c *ClusterCollector) getLastAppliedConfig(annotations []string) (string, bool) {
```

and then call it as:
```go
...
if jsonManifest, ok := c.getLastAppliedConfig(r.GetAnnotations()); ok {
```

I think it would make testing much easier.

stepanstipl: _It might still be easier for testing to refactor to:_
```go
func (c *ClusterCollector) getLastAppliedConfig(resourceAnnotations []string) (string, bool) {
```
_but either is fine._

pkg/collector/cluster.go
AndreasSko: Yeah, you are right. Good catch, thanks! :slightly_smiling_face:

pkg/config/config.go
```diff
@@ -34,6 +35,7 @@ func NewFromFlags() (*Config, error) {
 	}
 
 	flag.StringSliceVarP(&config.AdditionalKinds, "additional-kind", "a", []string{}, "additional kinds of resources to report in Kind.version.group.com format")
+	flag.StringSliceVar(&config.AdditionalAnnotations, "additional-annotations", []string{}, "additional annotations that should be checked to determine the last applied config")
```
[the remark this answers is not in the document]
AndreasSko: Good idea!

pkg/collector/cluster_test.go
AndreasSko: Yeah, that make sense :+1:

AndreasSko:
Thank you, @stepanstipl for your good review. I appreciate the feedback!
I adjusted the PR (hopefully I didn't forget anything :smile: :see_no_evil:) and it would be ready for another review :slightly_smiling_face:

README.md
```diff
@@ -76,19 +76,24 @@ You can list all the configuration options available using `--help` switch:
...
+  -k, --kubeconfig string               path to the kubeconfig file
+  -l, --log-level string                set log level (trace, debug, info, warn, error, fatal, panic, disabled) (default "info")
+  -o, --output string                   output format - [text|json] (default "text")
+  -t, --target-version string           target K8s version in SemVer format (autodetected by default)
+  -v, --version                         prints the version of kubent and exits
+
 ```
+- *`--additional-annotation`*
+  Check additional annotations for the last applied configuration. This can be useful if a resource was applied
+  with a tool other than kubectl. The flag can be used multiple times.
```
stepanstipl, line 96: Looks like `pre-commit` tests are failing, I think it's extra space at the end of this line.

stepanstipl (changes requested):
Looks good to me 👍 thanks!

I think it's only the trailing space, because of which the CI is failing, and we're good to merge.

README.md
AndreasSko: Hehe, sorry didn't check the CI. I removed the space now. Hopefully the pipeline gets green now :smile:

AndreasSko:
Thanks! Commit is signed now :slightly_smiling_face:

stepanstipl:
Thank you very much! 👍 This has been merged 🎉 

We're planning to do a release soon-ish (0.6.0, only a few bits left 🤞  [^1]), so it will be there. Or the next nightly 🌙 .

[^1]: https://github.com/doitintl/kube-no-trouble/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Minor+0.6.0%22

PR #329, opened 2022-08-03
feat: Cover deprecated autoscaling/v2beta1 API group - HPA

Jmspz:
Adding a scan for the HorizontalPodAutoscaler. API version autoscaling/v2beta1 was removed in 1.25 as well as the others here.


Jmspz:
Accidentally put both pieces in a single pull request. Will create two separate pull requests.

PR #331, opened 2022-08-03
feat: Add handling simple resource groups

Jmspz:
feat: Add handling simple resource groups when adding additional kinds to scan

   If the API version is not included in the code for resources in groups apps, batch,
   autoscaling, etc, add the ability to scan for those resources using the -a option.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -33,7 +33,7 @@ func NewFromFlags() (*Config, error) {
 		TargetVersion: &judge.Version{},
 	}
 
-	flag.StringSliceVarP(&config.AdditionalKinds, "additional-kind", "a", []string{}, "additional kinds of resources to report in Kind.version.group.com format")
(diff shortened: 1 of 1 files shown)


dark0dave:
I like this.

dark0dave (changes requested):
Adding flag requires readme/docs to be updated

stepanstipl:
@dark0dave this PR is not adding a flag, just fixing validation for the existing one.

@Jmspz thank you for this 👍

Can you please add an example here with what's currently not handled, and would be with this fix in place?

Internally we're using `ParseKindArg` [^1] so can that function handle this shorter form? I'm not 100% by the description, might need to be tested.

For sure we should add some tests for this.

[^1]: https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#ParseKindArg

PR #337, opened 2022-08-22
feat(golang19): update to golang 19, go mod tidy too

dark0dave:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,25 +1,70 @@
 module github.com/doitintl/kube-no-trouble
 
-go 1.16
+go 1.19
 
 require (
-	github.com/Azure/go-autorest/autorest/adal v0.9.5 // indirect
-	github.com/DATA-DOG/go-sqlmock v1.4.1 // indirect
 	github.com/ghodss/yaml v1.0.0
-	github.com/google/gofuzz v1.1.0 // indirect
 	github.com/hashicorp/go-version v1.3.0
-	github.com/jmoiron/sqlx v1.2.0 // indirect
-	github.com/lib/pq v1.3.0 // indirect
 	github.com/open-policy-agent/opa v0.34.2
-	github.com/paultyng/changelog-gen v1.1.0 // indirect
 	github.com/rs/zerolog v1.21.1-0.20210413053206-582f0cf0e39b
-	github.com/rubenv/sql-migrate v0.0.0-20200402132117-435005d389bc // indirect
 	github.com/spf13/pflag v1.0.5
 	helm.sh/helm v2.17.0+incompatible
 	helm.sh/helm/v3 v3.1.3
 	k8s.io/apimachinery v0.17.13
 	k8s.io/client-go v0.17.13
+)
+
+require (
+	cloud.google.com/go v0.81.0 // indirect
+	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
+	github.com/Azure/go-autorest/autorest v0.9.0 // indirect
+	github.com/Azure/go-autorest/autorest/adal v0.9.5 // indirect
+	github.com/Azure/go-autorest/autorest/date v0.3.0 // indirect
+	github.com/Azure/go-autorest/logger v0.1.0 // indirect
+	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
+	github.com/DATA-DOG/go-sqlmock v1.4.1 // indirect
+	github.com/Masterminds/semver/v3 v3.0.3 // indirect
+	github.com/OneOfOne/xxhash v1.2.8 // indirect
+	github.com/cyphar/filepath-securejoin v0.2.2 // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/evanphx/json-patch v4.9.0+incompatible // indirect
+	github.com/form3tech-oss/jwt-go v3.2.2+incompatible // indirect
+	github.com/gobwas/glob v0.2.3 // indirect
+	github.com/gogo/protobuf v1.3.2 // indirect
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -68,7 +68,7 @@ build: $(BINS)
 
 $(BIN_DIR)/%-$(BIN_ARCH)$(BIN_SUFFIX): $(SRC) go.mod go.sum
 	mkdir -p $(BIN_DIR)
-	$(GOBUILD) -ldflags="-s -w -X main.version=$(GIT_REF) -X main.gitSha=$(GIT_SHA)" \
+	$(GOBUILD) -buildvcs=false -ldflags="-s -w -X main.version=$(GIT_REF) -X main.gitSha=$(GIT_SHA)" \
```
stepanstipl, line 71: Is this causing any trouble? I'm not well familiar with the `buildvcs`flag, but I understand it should just include vcs info into the build. I'm not against disabling it, just that I would like to understand the reason.

go.mod
```diff
@@ -1,25 +1,70 @@
...
 	github.com/rs/zerolog v1.21.1-0.20210413053206-582f0cf0e39b
-	github.com/rubenv/sql-migrate v0.0.0-20200402132117-435005d389bc // indirect
 	github.com/spf13/pflag v1.0.5
 	helm.sh/helm v2.17.0+incompatible
 	helm.sh/helm/v3 v3.1.3
 	k8s.io/apimachinery v0.17.13
 	k8s.io/client-go v0.17.13
+)
+
+require (
```
stepanstipl, line 17: A couple of questions about these changes:

- Where do they come from? New go version?
- Is it a standard practice to have two `require` blocks? _(I see it's a split between explicit and indirect)_
- Why are there so many new `indirect` dependencies?

scripts/alpine-setup.sh
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env sh
 
 # Set strict error checking
-set -emou pipefail
+set -eou pipefail
```
stepanstipl, line 4: Is the `-m` option causing any trouble?

stepanstipl (commented):
Thanks @dark0dave 👍

I think it looks good, just would like to better understand the reason behind some of the changes before merging - see individual comments.

stepanstipl:
Btw. @dark0dave I have invited your new email to Snyk, have a look. I think the reason behind this fail are the new indirect dependencies - not sure what can we do about it (but I don;t know where are these coming from - see my Q. above)

Makefile
dark0dave: Yep, macos was refusing to build

scripts/alpine-setup.sh
dark0dave: job control disabled in docker image

dark0dave:
@stepanstipl I basically just ran go mod tidy

go.mod
dark0dave: New changes to go mod. Since go1.17

https://go.dev/doc/go1.17#go-command

dark0dave: > For the go command to correctly resolve transitive imports using the pruned module graph, the go.mod file for each module needs to include more detail about the transitive dependencies relevant to that module. If a module specifies go 1.17 or higher in its go.mod file, its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.)

dark0dave:
@stepanstipl let me know if I have resolved your questions.

go.mod
dark0dave: You can see example here: https://github.com/helm/helm/blob/main/go.mod

dark0dave: I feel these changes are good because we are explicit about which packages we are using.

stepanstipl: Yep, makes sense 👍

scripts/alpine-setup.sh
stepanstipl: Ok, we're not using it anyway, so I guess that's fine.

Makefile
stepanstipl: macOS binary in the GitHub CI? Or local? Do you have the error message?

I'm wondering what the underlying issue is, and if we can keep this enabled - I would prefer to resolve/understand the cause to introducing this.

dark0dave: @stepanstipl can browse the error here: https://github.com/doitintl/kube-no-trouble/runs/8050840335?check_suite_focus=true#step:5:83

stepanstipl: Looks like error with the safe directory (again: 🤦 ) - I've tried to add the git command go is using to get the vcsinfo directly in the build, see #344 and https://github.com/doitintl/kube-no-trouble/runs/8058460382?check_suite_focus=true#step:5:22, fails with the following:

```
fatal: detected dubious ownership in repository at '/__w/kube-no-trouble/kube-no-trouble'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/kube-no-trouble/kube-no-trouble
Error: Process completed with exit code 128.
```

I though this would be taken care of by using the `set-safe-directory: true` argument on the checkout action, but apparently not. I want to look into this as to why not.

dark0dave: yep, its the git issue again. The set safe-directory to true should do it.

dark0dave: @stepanstipl I seem to have fixed it! Have a look at https://github.com/doitintl/kube-no-trouble/pull/337/files#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3R55, I just ran the safe command again.

dark0dave: Might be due to different users per task?

stepanstipl (approved):
I think it's good to go 👍 .

Issue #325, opened 2022-07-21
User-agent


stepanstipl:
GKE now displays user-agent with deprecated K8S API calls [^1]; we should make sure that we set specific "kubent" user-agent header (not sure what is it at the moment), to allow for clear identification of kubent.

[^1]: https://cloud.google.com/kubernetes-engine/docs/deprecations/viewing-deprecation-insights-and-recommendations#view-deprecation-insights-recs

stepanstipl:
Looks like we have `kubent` in user-agent, but version info is incorrectly populated (`v0.0.0`)

dark0dave:
Fixed by #327

PR #338, opened 2022-08-22
feat: Display correct version in User-Agent HTTP header

dark0dave:

### DIFF
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -29,6 +29,10 @@ const (
 	EXIT_CODE_FOUND_ISSUES = 200
 )
 
+func generateUserAgent() string {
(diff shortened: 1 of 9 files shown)


dark0dave:
> I think it looks good in general +1 thanks.
> 
> However the `newKubeCollector` is also used in Helm v2/v3 collectors, so I assume the user agent version will need to be propagated there as well.
> 
> I would expect the tests to fail - maybe it needs rebase to catchup with some of the recent CI changes.

I will fix the tests and anything using the newKubeCollector should have the user agent set as its part of the Rest Client config.

stepanstipl (approved):
Looks good 👍 

FIxes: #325

PR #346, opened 2022-09-01
feat: Add support for detecting snapshot.storage.k8s.io/v1beta1

stepanstipl:
This PR adds support for detecting now deprecated [^1] `snapshot.storage.k8s.io/v1beta1` resources:
- `VolumeSnapshot`
- `VolumeSnapshotClass`
- `VolumeSnapshotContent`

Fixes https://github.com/doitintl/kube-no-trouble/issues/87

Other minor changes introduced:
- [test: Reduce test noise level](https://github.com/doitintl/kube-no-trouble/commit/78e9edfd9e289d3c63714ea5cf3b5b6aa9c9d58f): reduce log level in tests to `WARN` to improve readability of tests output

[^1]: https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/config/crd

### DIFF
--- a/fixtures/volumesnapshot-v1beta1.yaml
+++ b/fixtures/volumesnapshot-v1beta1.yaml
@@ -1,0 +1,8 @@
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshot
+metadata:
+  name: new-snapshot-demo
+spec:
+  volumeSnapshotClassName: csi-hostpath-snapclass
+  source:
+    persistentVolumeClaimName: hpvc
--- a/fixtures/volumesnapshotclass-v1beta1.yaml
+++ b/fixtures/volumesnapshotclass-v1beta1.yaml
@@ -1,0 +1,6 @@
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotClass
+metadata:
+  name: csi-hostpath-snapclass
+driver: hostpath.csi.k8s.io
+deletionPolicy: Delete
--- a/fixtures/volumesnapshotcontent-v1beta1.yaml
+++ b/fixtures/volumesnapshotcontent-v1beta1.yaml
@@ -1,0 +1,15 @@
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotContent
+metadata:
+  name: new-snapshot-content-test
+  annotations:
+    - snapshot.storage.kubernetes.io/allowVolumeModeChange: "true"
+spec:
+  deletionPolicy: Delete
+  driver: hostpath.csi.k8s.io
+  source:
+    snapshotHandle: 7bdd0de3-aaeb-11e8-9aae-0242ac110002
+  sourceVolumeMode: Filesystem
+  volumeSnapshotRef:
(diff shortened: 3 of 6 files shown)


stepanstipl:
_SonarCLoud is complaining about repeating `snapshot.storage.k8s.io` and suggesting a constant, but I don't think it's a problem in this case._

pkg/collector/collector_test.go
```diff
@@ -1,9 +1,14 @@
 package collector
 
 import (
+	"github.com/rs/zerolog"
 	"testing"
 )
 
+func init() {
```
dark0dave, line 8: does this belong here? Shouldn't this be in main?

test/helper.go
```diff
@@ -7,9 +7,14 @@ import (
 	"github.com/doitintl/kube-no-trouble/pkg/judge"
 	"github.com/doitintl/kube-no-trouble/pkg/rules"
 
+	"github.com/rs/zerolog"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+func init() {
```
dark0dave, line 14: if this is in a helper can we call it in the other packages?

dark0dave (changes requested):
I get the idea of having less noise but maybe we should have a dedicated package for tests to set noise level?

dark0dave:
Detection looks fine.

dark0dave:
Also if you want to change noise level please open another pr, feels separate to this.

test/helper.go
stepanstipl: Yep, we can, in the init? So what you're suggesting is to:

- refactor this as standard (non-init) function in this package
- call this function in the `init()` of this and other packages?

Basically, it needs to get called in once in every` _test` package. I'm open to reshuffling things around, just want to keep it simple.

stepanstipl:
Ok, fair enough, I will move that bit to another PR.

pkg/judge/rego_test.go
```diff
@@ -7,11 +7,16 @@ import (
 
 	"github.com/doitintl/kube-no-trouble/pkg/rules"
 	"github.com/ghodss/yaml"
+	"github.com/rs/zerolog"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 const FIXTURES_DIR = "../../fixtures"
 
+func init() {
```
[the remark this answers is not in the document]
stepanstipl: Moved to another PR

test/helper.go
stepanstipl: Moved to another PR

pkg/collector/collector_test.go
stepanstipl: Moved to another PR

dark0dave (approved):
Nice good work! fix: https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&pullRequest=346&id=doitintl_kube-no-trouble&open=AYL4l6JHTJesVHBCmzps

dark0dave (changes requested):
Also sign your commits @stepanstipl  :(

dark0dave (approved):
Good, thanks for this.

PR #350, opened 2022-09-01
chore: Users can install latest nightly

dark0dave:

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -13,6 +13,7 @@ GITHUB_REPO="${GITHUB_REPO:="doitintl/kube-no-trouble"}"
 TARGET_DIR="${TARGET_DIR:="/usr/local/bin"}"
 TARGET_ARCH="${TARGET_ARCH:="$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')"}"
 TARGET_OS="${TARGET_OS:="$(uname -s | tr '[:upper:]' '[:lower:]')"}"
+TARGET_NIGHTLY="${TARGET_NIGHTLY:=false}"
 REQUIRED_BINARIES=${REQUIRED_BINARIES:='tar curl'}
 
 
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -65,40 +67,51 @@ curl -sL "https://api.github.com/repos/${GITHUB_REPO}/releases/latest" \
...
          a) TARGET_ARCH="$OPTARG";;
          o) TARGET_OS="$OPTARG";;
+         n) TARGET_NIGHTLY="$OPTARG";;
          ?) usage; exit;;
      esac
 done
 
 echo "${YELLOW}>${NOCOL} Detecting latest version"
 release="$(get_latest_release)"
+if [ ${TARGET_NIGHTLY} -eq 1 ]; then
```
stepanstipl, line 112: Can we do if/else here? Otherwise, we're always downloading the latest tagged release info, even if that is not required.

stepanstipl: Also
```
In scripts/install.sh line 112:
if [ ${TARGET_NIGHTLY} -eq 1 ]; then
     ^---------------^ SC2086: Double quote to prevent globbing and word splitting.
```

scripts/install.sh
```diff
@@ -56,6 +57,7 @@ OPTIONS:
    -d      Directory where kubent will be installed. Default is /usr/local/bin
    -a      Architecture to install (x86_64 only atm.). Default is to auto-detect.
    -o      OS (linux, macos). Default is to auto-detect.
+   -n      Install latest nightly, if set to 1. Default is 0
```
stepanstipl, line 60: Can we make this boolean switch, instead of an argument with a value? (same as the -h)

I.e.:
```bash
./install.sh -n ...
```

as opposed to
```bash
./install.sh -n1 ...
```

stepanstipl (commented):
Good, some minor things in individual comments. Thanks 👍

scripts/install.sh
dark0dave: yep will fix

stepanstipl (approved):
Beautiful, thanks!

PR #355, opened 2022-09-08
Add support for configuring output file

### LINKED ISSUE #304: Add config option to allow output to specific file, e.g. `--output-file`

stepanstipl:
This PR adds support for `-O` flag to configure output to a file to main binary and all the existing printers (json, text).

Related changes introduced:
- Increased test coverage for the common code and existing printers in the `printer` package
- Refactored Main tests to pass arguments via env variable, this was required to maintain consistency between runs (tmp dir)

Fixes #304

_Sonar CLoud complains about test func naming with underscores (which are generated by Goland and questionably a standard for unexported func tests in Go) and some repeated strings all in tests. I plan to ignore it... in the longer term, we should really get to enabling some linters in a way that makes sense), the SC is getting annoying._

### DIFF
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -155,14 +155,9 @@ func main() {
 		log.Fatal().Err(err).Str("name", "Rego").Msg("Failed to filter results")
 	}
 
-	printer, err := printer.NewPrinter(config.Output)
+	err = outputResults(results, config.Output, config.OutputFile)
 	if err != nil {
-		log.Fatal().Err(err).Msg("Failed to create printer")
-	}
-
-	err = printer.Print(results)
-	if err != nil {
-		log.Fatal().Err(err).Msg("Failed to print results")
+		log.Fatal().Err(err).Msgf("Failed to output results")
 	}
 
 	if config.ExitError && len(results) > 0 {
@@ -171,4 +166,19 @@ func main() {
 		exitCode = EXIT_CODE_SUCCESS
 	}
 	os.Exit(exitCode)
+}
+
+func outputResults(results []judge.Result, outputType string, outputFile string) error {
+	printer, err := printer.NewPrinter(outputType, outputFile)
+	if err != nil {
+		return fmt.Errorf("failed to create printer: %v", err)
+	}
+	defer printer.Close()
+
+	err = printer.Print(results)
+	if err != nil {
+		return fmt.Errorf("failed to print results: %v", err)
+	}
+
+	return nil
 }
--- a/cmd/kubent/main_test.go
+++ b/cmd/kubent/main_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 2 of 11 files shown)


cmd/kubent/main.go
```diff
@@ -154,10 +154,11 @@ func main() {
 		log.Fatal().Err(err).Str("name", "Rego").Msg("Failed to filter results")
 	}
 
-	printer, err := printer.NewPrinter(config.Output)
+	printer, err := printer.NewPrinter(config.Output, config.OutputFile)
```
dark0dave, line 157: shouldnt this be only passed if the flag is set?

cmd/kubent/main.go
```diff
@@ -154,10 +154,11 @@ func main() {
...
 	if err != nil {
 		log.Fatal().Err(err).Msg("Failed to create printer")
 	}
+	defer printer.Close()
```
dark0dave, line 161: defer should be handled inside the printer

dark0dave: and only if the flag has been set

pkg/config/config.go
```diff
@@ -81,3 +91,20 @@ func validateAdditionalResources(resources []string) error {
...
 }
+
+// validateOutputFile checks if output file name is valid and if the
+// destination directory exists
+func validateOutputFile(outputFileName string) error {
+	if outputFileName == "" {
+		return fmt.Errorf("output file name can't be empty (use - for stdout)")
+	}
+
+	if outputFileName != "-" {
```
dark0dave, line 102: shouldn't this return and error? I would catch this above

if outputFileName == ""  or outputFileName == "-"

cmd/kubent/main.go
dark0dave, line 161: this should happen inside of printer

pkg/config/config.go
```diff
@@ -81,3 +91,20 @@ func validateAdditionalResources(resources []string) error {
 	}
 	return nil
 }
+
+// validateOutputFile checks if output file name is valid and if the
```
dark0dave, line 95: do we need the comments? I feel the name of the function is suffiecent

pkg/printer/printer.go
```diff
@@ -2,31 +2,75 @@ package printer
...
 	Print([]judge.Result) error
+	Close() error
 }
 
-func NewPrinter(choice string) (Printer, error) {
+type commonPrinter struct {
+	outputFile *os.File
+}
+
+// newCommonPrinter creates new printer that prints to given output file
```
dark0dave, line 24: these comments are not very useful

pkg/printer/printer.go
```diff
@@ -2,31 +2,75 @@ package printer
...
+func ParsePrinter(choice string) (func(string) (Printer, error), error) {
 	printer, exists := printers[choice]
 	if !exists {
 		return nil, fmt.Errorf("unknown printer type %s", choice)
 	}
 	return printer, nil
 }
+
+// ensureOutputFile will open file for writing, or create one if non-existent
+func ensureOutputFile(fileName string) (*os.File, error) {
```
dark0dave, line 65: ensureOutputFile ensureOutputFileExists

pkg/printer/printer.go
```diff
@@ -2,31 +2,75 @@ package printer
...
+	if fileName == "-" {
+		return os.Stdout, nil
+	}
+
+	of, err := os.OpenFile(fileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
dark0dave, line 70: does this work for windows?

dark0dave: might want to use ModeAppend instead

dark0dave (changes requested):
I think my main take away is that we should not be closing a file if we are using standard out, feels incorrect

cmd/kubent/main.go
stepanstipl: I would say no, given go doesn't really do optional args, we just pass the default (which is stdout `-`), same as we do just a couple of lines above, e.g. `loadedRules, err := rules.FetchRegoRules(additionalKinds)`.

cmd/kubent/main.go
stepanstipl: I don't know how to. Go doesn't have a concept of destructors, and `defer Close()` kind of a standard, see for example `io.Closer` [^1], and it's users responsibility to call it.

[^1]: https://pkg.go.dev/io#Closer

pkg/config/config.go
stepanstipl: Comments are standard for generating docs using `godoc`. And while it typically does not display unexported funcs, it does have a switch to do that. So for that reason, I would prefer to have, hopefully meaningful, comments as a standard.

I agree though that these are not very helpful, so would be happy for better suggestions :).

pkg/printer/printer.go
stepanstipl: Better suggestion? Explained above why I would like to have comments.

pkg/printer/printer.go
stepanstipl: This is commonly used in examples. I'm not sure though, haven't tested on Windows. Why shouldn't it?

pkg/config/config.go
dark0dave: ah, if they are generated, then we should like generate them for the whole repo. Rather than just these changes.

pkg/printer/printer.go
dark0dave: 0644 is a unix term

stepanstipl:
Thanks for reviewing.

> I think my main take away is that we should not be closing a file if we are using standard out, feels incorrect.

We're not, this is handled within the relevant printer's `Close()` implementation. In general, given lack of destructors in Go, "Closer" interface is a common pattern in go, e.g. see the previously mentioned `io.Closer` [^1], os `database/sql` package [^2].

But don't think about this just in the context of closing file, but as a generic destruct/cleanup interface to free any resources, and that any printer can implement as they wish. E.g. let's say we implement a "printer" to Cloud Logging, or to a database, this would be a place to teardown any open clients and connections.

[^1]: https://pkg.go.dev/io#Closer
[^2]: https://pkg.go.dev/database/sql@go1.19.1#Conn.Close

pkg/config/config.go
stepanstipl: I don't mean the comments themselves are generated, but that these are used in generated documentation. E.g. a few times mentions [^1] is generated from [^2]. I agree that at some time I would like to have comments for all funcs, probably when we put in some linter to check for that, but for now, I would prefer to add those rather than not.


[^1]: https://pkg.go.dev/io#Closer
[^2]: https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/io/io.go;l=100-106

cmd/kubent/main_test.go
```diff
@@ -211,3 +236,22 @@ func TestGetServerVersion(t *testing.T) {
 		t.Errorf("Expected %s version to be detected, instead got: %s", fakeVersion.String(), version.String())
 	}
 }
+
+func encodeBase64(args []string) (string, error) {
```
dark0dave, line 240: I think these are helpers?

dark0dave:
> Thanks for reviewing.
> 
> > I think my main take away is that we should not be closing a file if we are using standard out, feels incorrect.
> 
> We're not, this is handled within the relevant printer's `Close()` implementation. In general, given lack of destructors in Go, "Closer" interface is a common pattern in go, e.g. see the previously mentioned `io.Closer` [1](#user-content-fn-1-dd815946a879eb610c29af2e231f789d), os `database/sql` package [2](#user-content-fn-2-dd815946a879eb610c29af2e231f789d).
> 
> But don't think about this just in the context of closing file, but as a generic destruct/cleanup interface to free any resources, and that any printer can implement as they wish. E.g. let's say we implement a "printer" to Cloud Logging, or to a database, this would be a place to teardown any open clients and connections.
> ## Footnotes
> 
>     1. https://pkg.go.dev/io#Closer [leftwards_arrow_with_hook](#user-content-fnref-1-dd815946a879eb610c29af2e231f789d)
> 
>     2. https://pkg.go.dev/database/sql@go1.19.1#Conn.Close [leftwards_arrow_with_hook](#user-content-fnref-2-dd815946a879eb610c29af2e231f789d)

Sorry I worded that badly I don't mind defer.Close() at all, I am talking about closing the printer in main. Its not very obvious why we'd close the printer in main if we are not doing any file writing. Additionally its a different level of abstraction which probably should probably not be handled in main.

pkg/printer/printer.go
stepanstipl: I assume it should work just fine - https://pkg.go.dev/os#FileMode:
> A FileMode represents a file's mode and permission bits. The bits have the same definition on all systems, so that information about files can be moved from one system to another portably. Not all bits apply to all systems. ...

pkg/config/config.go
stepanstipl: That btw. brings up a very good question - we do build on many platforms, incl. Windows. Should we run tests on all of those as well?

 _I would say yes, but not sure how difficult is the implementation._

pkg/printer/printer.go
stepanstipl: That btw. brings up a very good question - we do build on many platforms, incl. Windows. Should we run tests on all of those as well?

I would say yes, but not sure how difficult is the implementation.

cmd/kubent/main_test.go
stepanstipl: Yep.

stepanstipl:
> Sorry I worded that badly I don't mind defer.Close() at all, I am talking about closing the printer in main. Its not very obvious why we'd close the printer in main if we are not doing any file writing. Additionally its a different level of abstraction which probably should probably not be handled in main.

Well, currently that's where the printer is created [^1]. I.e. typically `defer` follows after object creation, I'm not sure where else to put it (it also has to wait till we're done with the printer).

What would you suggest?

[^1]: https://github.com/doitintl/kube-no-trouble/blob/8c7bc37e1d3ae458327dfc420fef4397a1b016ff/cmd/kubent/main.go#L157

stepanstipl:
Renamed `ensureOutputFile` -> `ensureOutputFileExists` and moved printer code in main to its own function `outputResults` with tests.

@dark0dave I think it's ready for another look. Thanks.

dark0dave (approved):
LGTM

dark0dave:
@stepanstipl please rebase and fix conflicts

PR #359, opened 2022-09-16
chore: Add pre-commit tests to make test and add `test-ci` target

stepanstipl:
This is a simple PR to include pre-commit tests in make `test` target, and introduce a separate target (`test-ci`) for running tests in CI where tests covered by individual GH actions should be excluded.

This is a prep for introduction of linters, where it also makes sense to run those in individual GH action.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -30,7 +30,7 @@ jobs:
     - name: Test
       run: |
         scripts/alpine-setup.sh
-        make test
+        make test-ci
   build:
     name: build
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -80,10 +80,18 @@ $(RELEASE_DIR)/%-$(RELEASE_SUFFIX): $(BIN_DIR)/%-$(BIN_ARCH)$(BIN_SUFFIX)
 	mkdir -p $(RELEASE_DIR)
 	$(TAR) -cvz --transform 's,$(BIN_DIR)/$(*)-$(BIN_ARCH)$(BIN_SUFFIX),$(*)$(BIN_SUFFIX),gi' -f "$@" "$<"
 
-## Run Go tests
-test: test-fmt test-git
-	go test -v -coverprofile fmtcoverage.html ./...
+## Run all the tests
+test: test-fmt test-git test-pre-commit test-go
 .PHONY: test
+
+## Run all the tests in CI (excludes tests run via individual GH actions)
+test-ci: test-fmt test-git test-go
+.PHONY: test-ci
+
+## Run Go tests
+test-go:
+	go test -v -coverprofile fmtcoverage.html ./...
+.PHONY: test-go
 
 ## Run go and opt fmt checks
 test-fmt:
@@ -95,6 +103,11 @@ test-fmt:
 test-git:
 	./scripts/git-check-commits.sh
 .PHONY: test-git
+
+## Run pre-commit set of tests
+test-pre-commit:
+	pre-commit run --all-files
+.PHONY: test-pre-commit
 
 ## Clean build artifacts
 clean:


dark0dave (changes requested):
I think we should keep the tests together, happy to have a pre-commit step in the makefile tho

dark0dave:
issue is that, pre-commit stuff is in python so its not super clear, I really really hate the idea of having make ci test. We should have the same process run in either case. I would add the golang linters to the [pre-commit file](https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml). I think thats a better place for it rather than the make file.

stepanstipl:
> issue is that, pre-commit stuff is in python so its not super clear

I agree and would drop it in favour of makefile + all dependencies being in the docker container. But that's for another time.

> Consistency is king

Yes, `make test` should run a complete set of tests/checks. 

> the same process should run in either case

While I agree, I see the benefits of better integration with GH. So as I see it - yes, we should be able to execute all the tests locally/in the build docker env (`make test`), but as it has benefits it makes sense to run some of those via a dedicated step/action in CI. _Also currently we don't run the same - pre-commit is already using dedicated action + other bits in the GH build files are tightly coupled to GH._

> I would add the golang linters to the [pre-commit file](https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml). I think that is a better place for it rather than the make file.

I've explained the benefits of using the golang-ci action directly in previous post - i.e. speed (caching) and direct integration with GH PR. The pre-commit does not do this and currently is inconsistent as it's not executed as part of `make test`.

> Additionally I like separating out the linting and from the actual tests, as this keeps the speed of the tests reasonable. As you can see the tests are not as fast as the pre-commit:

Yes, this is done via several `test-*` make targets which both `test` and `test-ci` execute.

> which are run at the same time as the tests. Formatting I feel should be in its own stage, additionally I would like to add the git check to the pre-commit file. As this is the most logical place for it. Ie the commits should be checked before they are committed.

I think this depends on individual preferences a lot, but I for example don't run pre-commit on every commit. I simply want to run all the tests before pushing the commit/when I'm ready. With that logic you would want to run all the tests as pre-commit every time, but that would only further surface the issue -it's too slow. And we don't enforce all test/checks to pass on every commit (we could do that, but I feel that might be too much and would only further push towards doing large single-commits which is exactly opposite of what I prefer to encourage).

> PS I still hate the makefile

Please suggest a better alternative :), but again, separate topic.

dark0dave:
>I think this depends on individual preferences a lot, but I for example don't run pre-commit on every commit. I simply want to run all the tests before pushing the commit/when I'm ready. With that logic you would want to run all the tests as pre-commit every time, but that would only further surface the issue -it's too slow. And we don't enforce all test/checks to pass on every commit (we could do that, but I feel that might be too much and would only further push towards doing large single-commits which is exactly opposite of what I prefer to encourage).

I think you have misunderstood me. Pre-commits do not and should not run the test suite, only linting and commit message validation. You should run the pre-commit file against each commit, thats its purpose? And its not too slow check the actions. I disagree we should validate every commit where possible to ensure consistency. 


> Yes, this is done via several test-* make targets which both test and test-ci execute.

Again I don't think split these out into separate steps is good, make test is fine as is. Pre-commit is run per commit. You do not need to run make test to test commits. Seems backward no?

stepanstipl:
> And let me be clear you should run the pre-commit file against each commit, that is its purpose.

This is like you would be telling me I should only write code in VSCode and on Mac. 🤣

> I disagree we should validate every commit where possible to ensure consistency.

I'm potentially not against enforcing we move from one consistent state to another. I.e. that all the tests/checks must succeed for every commit - some projects/teams operate like this (although I think GH actions is not quite ready for this). Happy to discuss itfurther, but let's do that separately.

But I disagree with the logic we should run some tests only (the cheap ones, because they're cheap) for every commit locally (or better expect everyone to do so) and only run tests for the whole PR in CI (both the "pre-commit" ones and others).

> Again I don't think splitting these out into separate steps is good, make test is fine as is. pre-commit is run per commit. You do not need to run make test to test commits.

This is not correct - the CI runs pre-commit on the last commit from the PR only. Not on all commits.

So, perhaps to make this simpler. Can we agree that we should have one command to execute a complete suite of all the tests/checks?

stepanstipl:
I have plans to move the CI bit out of this project (as it's too much pain to change anything, in particular Github actions part always requires several attempts; is inconsistent and not reproducible locally and takes away focus from the core functionality of the project) to using a well-tested framework imported via git submodule.

Issue #357, opened 2022-09-13
Why does kubent scan kubectl.kubernetes.io/last-applied-configuration annotation instead of the actual apiVersion


hpl002:
This often leads to false positives.

stepanstipl:
I just commented in the other issue, this is misunderstanding of how K8S API works. 

**TL;DR:** because there is no such thing as the "actual" API version. The API version that the resource is returned in is dictated by the client, can be any of the supported versions by the given K8S API server version (i.e. you can get same resource in 2 different versions), and does not reflect how the resource was created.

> Hi everyone,
> 
> so kubent does look **only** at the `last-applied-configuration` (or Helm in-cluster manifests if you use Helm). This is described in [README](https://github.com/doitintl/kube-no-trouble) and I've tried to explain it in https://github.com/doitintl/kube-no-trouble/issues/276#issuecomment-1062082004. This is due to how K8s work, and basically the original version used to create the resource is not preserved anywhere else (leaving aside newer managed fields for now).
> 
> As to `kubectl` - K8S will return the resource in the API version that was requested by the client, please see https://github.com/doitintl/kube-no-trouble/issues/276#issuecomment-1062928787. This version is independent of a) the version used when the resource was created and b) of the internal stored version.
> 
> **TL;DR:** If the `kubectl.kubernetes.io/last-applied-configuration` annotation contains old API version, it's expected behaviour this will be returned by kubent. This is how kubent works and I would say it's desirable, as it suggests the resource was created using the old API version, i.e. you will run into trouble if you try to do the same after cluster upgrade.
> 
> I can see that all the examples shared here contain the old API version in the annotation, so I'm going to close this issue. But if that's not the case, please comment. Thanks 👍

Issue #310, opened 2022-06-27
Kubent returns deleted objects


jsirianni:
I have completely removed cert-manager but I still see a bunch of cert-manager objects

If I try to delete them, kubectl tells me it was not found. Im a bit stumped. For example:
```
kubectl delete clusterrolebinding -n cert-manager cert-manager
warning: deleting cluster-scoped resources, not scoped to the provided namespace
Error from server (NotFound): clusterrolebindings.rbac.authorization.k8s.io "cert-manager" not found
```

```
__________________________________________________________________________________________
>>> Deprecated APIs removed in 1.16 <<<
------------------------------------------------------------------------------------------
KIND         NAMESPACE      NAME                      API_VERSION    REPLACE_WITH (SINCE)
Deployment   cert-manager   cert-manager-cainjector   apps/v1beta1   apps/v1 (1.9.0)
Deployment   cert-manager   cert-manager-webhook      apps/v1beta1   apps/v1 (1.9.0)
Deployment   cert-manager   cert-manager              apps/v1beta1   apps/v1 (1.9.0)
__________________________________________________________________________________________
>>> Deprecated APIs removed in 1.22 <<<
------------------------------------------------------------------------------------------
KIND                             NAMESPACE      NAME                                                 API_VERSION                            REPLACE_WITH (SINCE)
APIService                       cert-manager   v1beta1.admission.certmanager.k8s.io                 apiregistration.k8s.io/v1beta1         apiregistration.k8s.io/v1 (1.10.0)
ClusterRoleBinding               cert-manager   cert-manager-cainjector                              rbac.authorization.k8s.io/v1beta1      rbac.authorization.k8s.io/v1 (1.8.0)
ClusterRoleBinding               cert-manager   cert-manager-webhook:auth-delegator                  rbac.authorization.k8s.io/v1beta1      rbac.authorization.k8s.io/v1 (1.8.0)
CustomResourceDefinition         <undefined>    scalingpolicies.scalingpolicy.kope.io                apiextensions.k8s.io/v1beta1           apiextensions.k8s.io/v1 (1.16.0)
CustomResourceDefinition         <undefined>    globalfelixconfigs.crd.projectcalico.org             apiextensions.k8s.io/v1beta1           apiextensions.k8s.io/v1 (1.16.0)
[...]

Issue #276, opened 2022-03-01
Ignore deprecated APIs in last-applied-configuration


mkemmerz:
kube-no-trouble also dectects deprecated APIs in configurations that are not directly applied at the moment. The example below should show this. kube-no-trouble detectes the in Kubernetes 1.22 depcreated API `networking.k8s.io/v1beta1` but the Ingress in running on a valid API `networking.k8s.io/v1`.

Therefore my feature request is to ignore deprecated API entries (and the warning) in `kubectl.kubernetes.io/last-applied-configuration` if the object is running on a non-deprecated API.

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress
  namespace: xxx
  uid: xxxx
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: >
      {"apiVersion":"networking.k8s.io/v1beta1","kind":"
```

davem-git:
I'm running in to this issue too. this took isn't helpful because of it

mkemmerz:
Unfortunately it seems like this is exactly what kube-no-trouble was designed for: https://github.com/doitintl/kube-no-trouble/blob/master/pkg/collector/cluster.go#L115

Can someone explain, maybe the author himself @stepanstipl, why it scans this configuration?

last-applied-configuration only represents the object state of the creation (see https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/):
```
This sets the kubectl.kubernetes.io/last-applied-configuration: '{...}' annotation on each object. 
The annotation contains the contents of the object configuration file that was used to create the object.
```

so if I make changes to the object after its creation kube-no-trouble will not be able to scan the new configuration and only detects outdated APIs.

davem-git:
say that again? the last applied version is what was applied before the latest one. We'll have a cluster upgrade to 1.22. shortly i'll respond if it causes any issues

mkemmerz:
@stepanstipl thanks a lot for the clarification! As this was the only documenttion I was able to find it was a misunderstanding from my side 😄 

@davem-git good luck! I am also interested if you ran into any problems. We are going to upgrade soon too.

As this ticket makes no sense now I will close.

stepanstipl:
@davem-git Not quite - what I was trying to say is that the API version of how the resource is created is not related to how it's stored internally, and how it's retrieved by the client.

Let me give a little example:

[...]

PR #370, opened 2022-09-27
refactor: use a struct instead of map[string]interface{}

luisdavim:
* Use a `struct` instead of `map[string]interface{}`
* Use the actual object instead of the last applied configuration annotation

~~If you think checking the `kubectl` annotation is important, I can add it as an extra check but I do think we should check the actual object first~~
**update:** I've added a new flag to toggle checking the `kubectl` annotation

Note: I had to move `Version` from the `judge` package to the `collector` package to avoid cyclic dependencies, maybe version should be its own package?

I've tested a build from my branch against a real EKS cluster with version 1.23:

```console
$ ./bin/kubent-darwin-amd64  
8:06PM INF >>> Kube No Trouble `kubent` <<<
8:06PM INF version dev (git sha dev)
8:06PM INF Initializing collectors and retrieving data
8:06PM INF Target K8s version is 1.23.10-eks-15b7512
8:06PM INF Retrieved 930 resources from collector name=Cluster
8:06PM INF Retrieved 0 resources from collector name="Helm v2"
8:07PM INF Retrieved 157 resources from collector name="Helm v3"
8:07PM INF Loaded ruleset name=custom.rego.tmpl
8:07PM INF Loaded ruleset name=deprecated-1-16.rego
8:07PM INF Loaded ruleset name=deprecated-1-22.rego
8:07PM INF Loaded ruleset name=deprecated-1-25.rego
8:07PM INF Loaded ruleset name=deprecated-future.rego
>>> Deprecated APIs removed in 1.25 <<<
KIND                  NAMESPACE     NAME                                        API_VERSION      REPLACE_WITH (SINCE)
PodDisruptionBudget   kube-system   cluster-autoscaler-aws-cluster-autoscaler   policy/v1beta1   policy/v1 (1.21.0)
PodDisruptionBudget   kube-system   ingress-nginx-controller                    policy/v1beta1   policy/v1 (1.21.0)
PodDisruptionBudget   kube-system   ingress-nginx-internal-controller           policy/v1beta1   policy/v1 (1.21.0)
PodDisruptionBudget   kyverno       kyverno                                     policy/v1beta1   policy/v1 (1.21.0)
PodSecurityPolicy     <undefined>   eks.privileged                              policy/v1beta1   <removed> (1.21.0)
PodSecurityPolicy     <undefined>   pl                                          policy/v1beta1   <removed> (1.21.0)
```

### DIFF
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -33,8 +33,8 @@ func generateUserAgent() string {
 	return fmt.Sprintf("kubent (%s/%s)", version, gitSha)
 }
 
-func getCollectors(collectors []collector.Collector) []map[string]interface{} {
-	var inputs []map[string]interface{}
+func getCollectors(collectors []collector.Collector) []collector.MetaOject {
+	var inputs []collector.MetaOject
 	for _, c := range collectors {
 		rs, err := c.Get()
 		if err != nil {
@@ -81,7 +81,7 @@ func initCollectors(config *config.Config) []collector.Collector {
 	return collectors
 }
 
-func getServerVersion(cv *judge.Version, collectors []collector.Collector) (*judge.Version, error) {
+func getServerVersion(cv *collector.Version, collectors []collector.Collector) (*collector.Version, error) {
 	if cv == nil {
 		for _, c := range collectors {
 			if versionCol, ok := c.(collector.VersionCollector); ok {
@@ -121,6 +121,9 @@ func main() {
 	initCollectors := initCollectors(config)
 
 	config.TargetVersion, err = getServerVersion(config.TargetVersion, initCollectors)
+	if err != nil {
+		log.Warn().Msgf("failed to get server version: %s", err)
+	}
 	if config.TargetVersion != nil {
 		log.Info().Msgf("Target K8s version is %s", config.TargetVersion.String())
 	}
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -27,6 +27,7 @@ const (
 	EXIT_CODE_SUCCESS      = 0
 	EXIT_CODE_FAIL_GENERIC = 1
 	EXIT_CODE_FOUND_ISSUES = 200
+	LastAppliedConfig      = "kubectl.kubernetes.io/last-applied-configuration"
 )
 
 func generateUserAgent() string {
@@ -110,6 +111,10 @@ func main() {
 	if config.KubentVersion {
 		log.Info().Msgf("version %s (git sha %s)", version, gitSha)
 		os.Exit(EXIT_CODE_SUCCESS)
+	}
+
+	if config.LastAppliedConfig {
+		config.AdditionalAnnotations = append(config.AdditionalAnnotations, LastAppliedConfig)
 	}
 
 	zerolog.SetGlobalLevel(zerolog.Level(config.LogLevel))
--- a/cmd/kubent/main_test.go
+++ b/cmd/kubent/main_test.go
@@ -231,7 +231,7 @@ func TestGetServerVersion(t *testing.T) {
 		t.Errorf("Failed to get version with error: %s", err)
 	}
 
-	fakeVersion, _ := judge.NewVersion(collector.FAKE_VERSION)
+	fakeVersion, _ := collector.NewVersion(collector.FAKE_VERSION)
 	if version.Compare(fakeVersion.Version) != 0 {
 		t.Errorf("Expected %s version to be detected, instead got: %s", fakeVersion.String(), version.String())
 	}
@@ -257,7 +257,7 @@ func decodeBase64(dst *[]string, encoded string) error {
 }
 
 func Test_outputResults(t *testing.T) {
-	testVersion, _ := judge.NewVersion("4.5.6")
+	testVersion, _ := collector.NewVersion("4.5.6")
 	testResults := []judge.Result{{"name", "ns", "kind",
 		"1.2.3", "rs", "rep", testVersion}}
 
--- a/fixtures/volumesnapshotcontent-v1beta1.yaml
+++ b/fixtures/volumesnapshotcontent-v1beta1.yaml
@@ -3,7 +3,7 @@ kind: VolumeSnapshotContent
 metadata:
   name: new-snapshot-content-test
   annotations:
-    - snapshot.storage.kubernetes.io/allowVolumeModeChange: "true"
+    snapshot.storage.kubernetes.io/allowVolumeModeChange: "true"
 spec:
   deletionPolicy: Delete
   driver: hostpath.csi.k8s.io
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	helm.sh/helm/v3 v3.1.3
 	k8s.io/apimachinery v0.17.13
 	k8s.io/client-go v0.17.13
+	sigs.k8s.io/yaml v1.2.0
 )
 
 require (
@@ -66,5 +67,4 @@ require (
 	k8s.io/klog v1.0.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20200410145947-bcb3869e6f29 // indirect
 	k8s.io/utils v0.0.0-20200324210504-a9aa75ae1b89 // indirect
-	sigs.k8s.io/yaml v1.2.0 // indirect
 )
--- a/pkg/collector/cluster.go
+++ b/pkg/collector/cluster.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/restmapper"
+	"sigs.k8s.io/yaml"
 )
 
 const CLUSTER_COLLECTOR_NAME = "Cluster"
@@ -125,19 +126,19 @@ func (c *ClusterCollector) Get() ([]MetaOject, error) {
 				return nil, err
 			}
 			results = append(results, manifest)
+			if manifest.Annotations == nil {
+				continue
+			}
+			for _, annotation := range c.additionalAnnotations {
+				if config, ok := manifest.Annotations[annotation]; ok {
+					var m MetaOject
+					if err := yaml.Unmarshal([]byte(config), &m); err == nil {
+						results = append(results, m)
+					}
+				}
+			}
 		}
 	}
 
 	return results, nil
-}
-
-func (c *ClusterCollector) getLastAppliedConfig(resourceAnnotations map[string]string) (string, bool) {
-	annotations := append([]string{"kubectl.kubernetes.io/last-applied-configuration"}, c.additionalAnnotations...)
-	for _, annotation := range annotations {
-		if jsonManifest, ok := resourceAnnotations[annotation]; ok {
-			return jsonManifest, ok
-		}
-	}
-
-	return "", false
 }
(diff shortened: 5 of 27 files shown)


stepanstipl:
Hi @luisdavim, thanks for this 👍.

First, there are a bit too many things mixed in one PR. Due to that I won't review everything, but will comment on the main points:

> Use the actual object instead of the last applied configuration annotation

This is a bit of a fundamental misconception of how K8S API, and subsequently _kubent_, works. The version in the resource itself can't be used as it's governed by the client. I have explained this a few times in the tickets, please see [^1] for more details and further links. 

> update: I've added a new flag to toggle checking the kubectl annotation

You can disable the cluster collector, using the -`c=false` flag, but having a cluster collector without checking the last-applied annotation wouldn't make any sense.

> Use a struct instead of map[string]interface{}

I'm failing to see benefits of this change - i.e. we're introducing a custom struct, which is not equivalent. The `map[string]interface{}` would hold the whole object, while the new struct would just hold metadata.

Also at a quick first look, I've noticed some issues with the implementation, such as dropping support for parsing JSON files, or replacing the yaml library.

Having said that, and while I'm not a fan of the major changes above (or perhaps I don't understand the benefits in case of the struct/replacing the yaml library), some of the code cleanups look nice, and I would be happy to accept those. But the main point would be please open a separate PR for each type of change that you think is desireable, so that we can discuss and review it individually, without being entangled with all the other changes.

Cheers 👍 


[^1]: https://github.com/doitintl/kube-no-trouble/issues/357#issuecomment-1245338759

luisdavim:
Hi, thanks for the feedback, some notes:

* I didn't drop support for `JSON` I've switched to the k8s `YAML` lib that handles both by first converting yaml to json and then do the marsheling/unmarsheling
* It's true that by using `map[string]interface{}` you get the full object, but that's not being used so using a struct to capture what we actually need from the object makes the code cleaner and simpler
* The flag is to disable the use of the `kubectl` annotation and use the actual object
* A big advantage of this change is also that allowed displaying the collector name, so users don't get confused when the tool complains about some resource that has been updated or doesn't even exist like in #310 I know about how objects are stored in `etcd` but if they can be retrieved with the supported version they are not a problem, that means they were converted, this is why it's important to check both and print the source (collector) so the user can identify false positives, having an object that is using a supported API but has the annotation with the unsupported API tells the users they should check the source file but if they don't know the source they just get confused.
* I can break this into separate PRs if you are still interested in this.

Here's an example output from my latest commit:

```console
$ go run cmd/kubent/main.go                                    
4:57PM INF >>> Kube No Trouble `kubent` <<<
4:57PM INF version dev (git sha dev)
4:57PM INF Initializing collectors and retrieving data
4:57PM INF Target K8s version is 1.23.10-eks-15b7512
4:57PM INF Retrieved 1022 resources from collector name=Cluster
4:57PM INF Retrieved 0 resources from collector name="Helm v2"
4:57PM INF Retrieved 157 resources from collector name="Helm v3"
4:57PM INF Loaded ruleset name=custom.rego.tmpl
4:57PM INF Loaded ruleset name=deprecated-1-16.rego
4:57PM INF Loaded ruleset name=deprecated-1-22.rego
4:57PM INF Loaded ruleset name=deprecated-1-25.rego
4:57PM INF Loaded ruleset name=deprecated-future.rego
>>> Deprecated APIs removed in 1.25 <<<
KIND                  NAMESPACE     NAME                                        API_VERSION      REPLACE_WITH (SINCE)   COLLECTOR
PodDisruptionBudget   kube-system   cluster-autoscaler-aws-cluster-autoscaler   policy/v1beta1   policy/v1 (1.21.0)     Helm v3
PodDisruptionBudget   kube-system   ingress-nginx-controller                    policy/v1beta1   policy/v1 (1.21.0)     Helm v3
PodDisruptionBudget   kube-system   ingress-nginx-internal-controller           policy/v1beta1   policy/v1 (1.21.0)     Helm v3
PodDisruptionBudget   kyverno       kyverno                                     policy/v1beta1   policy/v1 (1.21.0)     Helm v3
PodSecurityPolicy     <undefined>   eks.privileged                              policy/v1beta1   <removed> (1.21.0)     Cluster-annotation
PodSecurityPolicy     <undefined>   eks.privileged                              policy/v1beta1   <removed> (1.21.0)     <undefined>
PodSecurityPolicy     <undefined>   pl                                          policy/v1beta1   <removed> (1.21.0)     Cluster
```

stepanstipl:
Hi @luisdavim,

thanks for getting back and clarifying some of the points. I will comment only on the core problem:

> The flag is to disable the use of the kubectl annotation and use the actual object
> 
> A big advantage of this change is also that allowed displaying the collector name, so users don't get confused when the tool complains about some resource that has been updated or doesn't even exist like in https://github.com/doitintl/kube-no-trouble/issues/310 I know about how objects are stored in etcd but if they can be retrieved with the supported version they are not a problem, that means they were converted, this is why it's important to check both and print the source (collector) so the user can identify false positives, having an object that is using a supported API but has the annotation with the unsupported API tells the users they should check the source file but if they don't know the source they just get confused.

This is a misconception of how K8S versioning works, and this change doesn't make sense. As I've explained[^1], and more in detail[^2], the returned version is governed by the client and not stored anywhere. I.e. you can get the same object in two versions:

> ...
> $ kubectl get ingress.v1beta1.networking.k8s.io test -o yaml
> Warning: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
> apiVersion: networking.k8s.io/v1beta1
> kind: Ingress
> metadata:
> ...
> 
> and also
> 
> $ kubectl get ingress.v1.networking.k8s.io test -o yaml
> apiVersion: networking.k8s.io/v1
> kind: Ingress
> metadata:
>  ...

I.e. should we only look at the `apiVersion` of the returned object, it will be governed by the API version we request[^3], and never get anything deprecated. I.e. the cluster collector would never return anything and therefore be a bit pointless.

[^1]: https://github.com/doitintl/kube-no-trouble/issues/276#issuecomment-1062082004
[^2]: https://github.com/doitintl/kube-no-trouble/issues/276#issuecomment-1062928787
[^3]: https://github.com/doitintl/kube-no-trouble/blob/master/pkg/collector/cluster.go#L73-L107

luisdavim:
@stepanstipl I don't think you understood what I meant. 
The api version conversion is not done on the client-side. See https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning

> API resources are distinguished by their API group, resource type, namespace (for namespaced resources), and name. The API server handles the conversion between API versions transparently: all the different versions are actually representations of the same persisted data. The API server may serve the same underlying data through multiple API versions.

Users use this tool to check if they have resources in the cluster that could be a problem, not all the resources will be applied by `kubectl` or helm, some might be generated by a controller. Checking both the actual object and the annotations and reporting where the violation comes from (actual object, helm release, annotation, etc..) allows the user to understand if the issue is a false positive as well has the source, foe example, you can fix the latest helm release but know you can't rollback because the resources in a previous release are not supported, or maybe a resource was patched to fix the issue but the kubectl annotation was not updated, or the resource is managed by some other tool but at some point was edited with kubectl so it got an annotation that may now be stale.

What I'm doing here is to check both and let the user know which collector reported the issue, if the issue comes from an annotation instead of the actual object the source will be listed as `Cluster-annotation`

For example, the issue that got me here was a resource that was managed by EKS that for some reason had a stale kubectl annotation, this change would still warn about it but make the debugging easier, I had to read the tool's code to understand why this was being reported as an issue.

stepanstipl:
Hi @luisdavim please check the above example with Ingress. The *same* object will always be returned in the version requested by the client. I.e. same object -> 2 different API versions.

> The API server may serve the same underlying data through multiple API versions.

The API version of the returned object will be governed by our client and checking against those has no meaning as the deprecated versions would only be returned if we would have requested those in the first place (and we do not [^1]).

This is better described in [^2], but there is ever only one internal version of a given resource. However, resources can be created in any of the supported API versions and conversion to internal format will happen. And the other way when retrieving the resource. But again, the returned API version is governed by _kubent_ as a client, i.e. we get what we request and checking against that makes no sense.

I understand this can be confusing for users, but unfortunately, that's how K8S API works in the first place. _Some of the usability improvements such as displaying the collector column are potentially nice, and we can discuss those separately._

 _If it would be as simple as checking the returned `apiVersion`, this tool would not exist in the first place. 😉_

[^1]: https://github.com/doitintl/kube-no-trouble/blob/master/pkg/collector/cluster.go#L73-L107 [↩](https://github.com/doitintl/kube-no-trouble/pull/370#user-content-fnref-3-59686eb3c7ba9a0c57b947fd29e94c65)
[^2]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md

luisdavim:
Yes @stepanstipl , I get that. As I've said, I understand how the resources are being stored in `etcd` and retreived fom the API but there are edge cases that the tool doesn't cover.

 If the API can convert the resrouce into the new version it won't be a problem and will not be reported unless the `kubectl` annotation has the deprecated API version. What I'm trying to cover with these changes are cases like the following:
* The resouce doesn't have a `kubectl` annotation and it's not part of any helm release and the API can't convert it, for example if you have some controller that generates PSPs (a resource that will be removed), without these changes that issue will not be reported.
* The resource can be converted and it's not really a problem but the `kubectl` annotation is referring to a deprecated API, in this case the tool shhould (and still wiil with these changes) report the issue but also let the user know it comes from the annotation and not the resource itself.

The point is, I think these changes make the tool more roubust and its output more informative. I'm not removing any functionality or behaviour, only making the tool also check the actual object and letting the user know which collector reported the issue so a better informed decision can be made.

luisdavim:
Also note that I'm removing more code than I'm adding whilst sill adding and not removing functionality...

stepanstipl:
> * Use the actual object ~~instead of~~ and the last applied configuration annotation
> ~~If you think checking the `kubectl` annotation is important, I can add it as an extra check but I do think we should check the actual object first~~

So this PR started as a suggestion to replace the annotation, with checking the returned API version directly, which as explained, does not work.

> The resouce doesn't have a kubectl annotation and it's not part of any helm release and the API can't convert it, for example if you have some controller that generates PSPs (a resource that will be removed), without these changes that issue will not be reported.

There's only one case - resource - currently where this would work (PSP). Conceptually I'm against implementing behaviour that introduces special handling for individual resources. It's more confusing than beneficial IMHO.

> The resource can be converted and it's not really a problem but the kubectl annotation is referring to a deprecated API, in this case the tool shhould (and still wiil with these changes) report the issue but also let the user know it comes from the annotation and not the resource itself.

Improving output is potentially a good feature. Atm. I'm not convinced this should be done by changing the default one, as the collector is usually not something the user should be concerned about, but perhaps the output can be made more configurable (e.g. specify what columns should be printed). It would be for a discussion on how to expose this to the user - but please via separate issue/PR. Discussing all this together is difficult.

> Also note that I'm removing more code than I'm adding whilst sill adding and not removing functionality...

_This is mainly caused by removed tests, which are not adequately replaced (e.g. no tests for `parseManifest()`). 😉 And there are a few other implementation issues._

To sum up - the core change of this PR is IMHO misconception of how K8S API and kubent work and not desirable. I would be happy to discuss other proposed changes (already commented on most of those), but please let's do this via small targeted Issues/PRs so that it's more feasible to have a focused conversation about and review each of those. I'm not going to respond to this PR anymore to protect my time. Thanks.

luisdavim:
This would bot be to cover the PSP specifically but to cover the case of removed APIs.

PR #373, opened 2022-09-30
fix: Helm - only default to release namespace for namespaced resources

### LINKED ISSUE #195: Helm - namespace incorrectly assigned for Helm cluster-level resources
We incorrectly assign Chart namespace even for Cluster-level resources. These should not have a namespace:

```
>>> Deprecated APIs removed in 1.22 <<<
KIND                       NAMESPACE     NAME                                             API_VERSION                         REPLACE_WITH (SINCE)
...
...
StorageClass               default       slow                                             storage.k8s.io/v1beta1              storage.k8s.io/v1 (1.6+)
```

stepanstipl:
This uses discovery client to check if given type of resource is namespaced. Discovery client should be cached, so it's fine to call ServerResourcesForGroupVersion() multiple times.

Fixes #195

### DIFF
--- a/pkg/collector/helm.go
+++ b/pkg/collector/helm.go
@@ -2,12 +2,16 @@ package collector
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
 
 	"github.com/ghodss/yaml"
+	"github.com/rs/zerolog/log"
 	"helm.sh/helm/v3/pkg/releaseutil"
 )
 
-func parseManifests(manifest string, defaultNamespace string) ([]map[string]interface{}, error) {
+func parseManifests(manifest string, defaultNamespace string, discoveryClient discovery.DiscoveryInterface) ([]map[string]interface{}, error) {
 	var results []map[string]interface{}
 
 	manifests := releaseutil.SplitManifests(manifest)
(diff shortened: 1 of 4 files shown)


dark0dave:
I think the sonar change is valid: Rename function "Test_isResourceNamespaced" to match the regular expression ^(_|[a-zA-Z0-9]+)$

pkg/collector/helm.go
```diff
@@ -2,12 +2,16 @@ package collector
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
dark0dave, line 5: not sure about the apimachinery library

pkg/collector/helm.go
```diff
@@ -20,22 +24,39 @@ func parseManifests(manifest string, defaultNamespace string) ([]map[string]inte
...
 
-		results = append(results, manifest)
+		results = append(results, fixed)
 	}
 
 	return results, nil
 }
 
-func fixNamespace(manifest *map[string]interface{}, defaultNamespace string) {
+func fixNamespace(manifest map[string]interface{}, defaultNamespace string, discoveryClient discovery.DiscoveryInterface) map[string]interface{} {
```
dark0dave, line 35: I would prefer the pointer, can't we use the unstructured. can't we deference instead? for 

```
&unstructured.Unstructured{Object: *manifest}
```

dark0dave (changes requested):
Just some small minor changes, looks good other wise

pkg/collector/helm.go
dark0dave: If its used else where ignore me

dark0dave (approved):

Issue #378, opened 2022-10-04
Binary inside docker image has incorrect version


stepanstipl:
```sh
$ docker run -it --rm ghcr.io/doitintl/kube-no-trouble:nightly-0.5.1-75-g841b420 -v
11:55AM INF version dev (git sha dev)
```

stepanstipl:
Did not work 😢 :
```
$ docker run -it --rm ghcr.io/doitintl/kube-no-trouble:nightly-0.5.1-80-gabba03b -v
Unable to find image 'ghcr.io/doitintl/kube-no-trouble:nightly-0.5.1-80-gabba03b' locally
nightly-0.5.1-80-gabba03b: Pulling from doitintl/kube-no-trouble
464c008b0010: Pull complete
b150c4772e4b: Pull complete
Digest: sha256:48c84abe67a7873907209d98dbd25989870152a51d1eb9ddae94f1f63838d2f8
Status: Downloaded newer image for ghcr.io/doitintl/kube-no-trouble:nightly-0.5.1-80-gabba03b
10:23AM INF version dev (git sha dev)
```

PR #382, opened 2022-10-05
ci: Fix missing build info in docker images

stepanstipl:
Any Docker build arguments must be explicitly declared in the relevant section of Dockerfile [^1], otherwise, these are not visible to build inside even when passed via `--build-arg`.

Tested locally:

```shell
$ GITHUB_REF=refs/tags/xxx GITHUB_SHA=123 docker build --build-arg GITHUB_REF --build-arg GITHUB_SHA -t test ./
...
...
=> [builder 11/11] RUN make all                                                                                                                                               25.6s
 => => # mkdir -p bin
 => => # go build -ldflags="-s -w -X main.version=xxx -X main.gitSha=123" \
...
...
$ docker run -it --rm test -v                                                                                                                           0
11:03AM INF version xxx (git sha 123)
```

[^1]: https://docs.docker.com/engine/reference/builder/#arg

Fixes: #378

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 FROM golang:1.19.1-alpine3.16 as builder
+ARG GITHUB_REF GITHUB_SHA
 WORKDIR /src
 COPY go.mod go.sum ./
 COPY scripts scripts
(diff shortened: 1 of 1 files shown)


dark0dave:
I have built this too seems good.

dark0dave (approved):

PR #383, opened 2022-10-06
ci: Fix latest tag and avoid rebuild

stepanstipl:
This should, hopefully 🤞, avoid rebuilding of the docker image with a different hash (and also currently broken as it's missing the build-args).

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -169,17 +169,12 @@ jobs:
         with:
           context: .
           push: true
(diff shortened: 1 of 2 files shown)


dark0dave:
@stepanstipl its not a rebuild as the image layers should be cached.

dark0dave:
This only pushes one tag? And then latest would be overwritten by nightlies? I thought we want to avoid that?

dark0dave (approved):
This feels more painful than it should.

stepanstipl:
Yep, it definitely does. Hands up for CI that allows you to test changes locally 🤣

PR #394, opened 2022-10-28
ci: Fix latest tag issue in nightly builds

stepanstipl:
Another try to fix broken tags on docker building GH actions.

Currently nightly builds are broken, as the `${{ !startsWith(github.ref, 'refs/tags/nightly') && format('{0}/{1}:{2}', env.REGISTRY, env.IMAGE_NAME, 'latest') }}` evaluates to `false`. After this change, it will evaluate to an empty string.

_Warning: not 100% sure this will work - while the syntax will return an empty string, I'm not 100% sure if the `docker/build-push-action@v3` action will deal with that. If this won't work, we can try the metadata action[^1], but that seems pretty complex too._

[^1]: https://github.com/docker/metadata-action

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -171,7 +171,7 @@ jobs:
           push: true
           tags: |
             ${{ format('{0}/{1}:{2}', env.REGISTRY, env.IMAGE_NAME, needs.create-release.outputs.tag_name) }}
-            ${{ !startsWith(github.ref, 'refs/tags/nightly') && format('{0}/{1}:{2}', env.REGISTRY, env.IMAGE_NAME, 'latest') }}
+            ${{ ( !startsWith(github.ref, 'refs/tags/nightly') && format('{0}/{1}:{2}', env.REGISTRY, env.IMAGE_NAME, 'latest') ) || '' }}
           build-args: |
             GITHUB_REF
             GITHUB_SHA
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.SHELL := bash
+SHELL := bash
 .SHELLFLAGS := -eu -o pipefail -c
 .ONESHELL:
 .SECONDEXPANSION:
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yaml
```diff
@@ -171,7 +171,7 @@ jobs:
           push: true
           tags: |
             ${{ format('{0}/{1}:{2}', env.REGISTRY, env.IMAGE_NAME, needs.create-release.outputs.tag_name) }}
-            ${{ !startsWith(github.ref, 'refs/tags/nightly') && format('{0}/{1}:{2}', env.REGISTRY, env.IMAGE_NAME, 'latest') }}
+            ${{ !startsWith(github.ref, 'refs/tags/nightly') && format('{0}/{1}:{2}', env.REGISTRY, env.IMAGE_NAME, 'latest') || '' }}
```
dark0dave, line 174:
```suggestion
            ${{ ( !startsWith(github.ref, 'refs/tags/nightly') && format('{0}/{1}:{2}', env.REGISTRY, env.IMAGE_NAME, 'latest') ) || ''  }}
```

dark0dave: lets try to be exact

dark0dave (changes requested):
Thanks,

.github/workflows/main.yaml
stepanstipl: It should be exactly the same given `&&` operator has precedence over `||` operator (same with the `!` op.), but I will add the brackets for readability.

dark0dave (approved):

dark0dave:
@stepanstipl I'll fixup the issue here, I am pretty sure I know what it is

dark0dave:
@stepanstipl should be fine now.

stepanstipl:
👍 thanks for the missing bash fix... I hope this will also fix the issue on your Linux distro...

PR #436, opened 2023-01-17
add option for DoNotResolvePaths

artificial-aidan:
Signed-off-by: Aidan Jensen <aidan@artificial.com>

### DIFF
--- a/pkg/collector/cluster.go
+++ b/pkg/collector/cluster.go
@@ -24,15 +24,16 @@ type ClusterCollector struct {
 }
 
 type ClusterOpts struct {
-	Kubeconfig      string
-	KubeContext     string
-	ClientSet       dynamic.Interface
-	DiscoveryClient discovery.DiscoveryInterface
+	Kubeconfig            string
+	KubeContext           string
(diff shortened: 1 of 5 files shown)


stepanstipl:
Hi @artificial-aidan, thanks for this. Before getting to a more technical review, I would like to understand the change itself.

Mainly, can you please elaborate a bit on what this change is doing and what problem it solves? 



And a couple of additional questions:
1) It seems this would be set to `true` by default, i.e. it would change the current behaviour by default?
```go
	flag.BoolVar(&config.DoNotResolvePaths, "do-not-resolve-paths", true, "kubeconfig loader setting \"DoNotResolvePaths\"")
```
2) The go-client code [^1] seems to suggest the `false`, i.e. to resolve paths, is usually desirable?
3) How does `kubectl` behave here by default? And is this configurable?

_(I believe that to mirror `kubectl`'s behaviour, at least by default, is what we want in most cases)._


Thanks


[^1]: https://github.com/kubernetes/client-go/blob/dda9c5244f2d0157d401974a53deb8154a2184cd/tools/clientcmd/loader.go#L122

PR #461, opened 2023-03-08
Add missing buildx setup action for image build

kkopachev:
I noticed that [latest build failed](https://github.com/doitintl/kube-no-trouble/actions/runs/4258761344/jobs/7410371441) to push docker images because docker buildx was not set up properly

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -177,6 +177,8 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


dark0dave:
I think this change has already been merged. Thank you. I will close this.

dark0dave:
> @dark0dave has it? Can you please link the PR/commit?
> 
> Unless I'm missing smth. this is still needed (this would only manifest when tag is pushed and the relevant part of CI build runs).

Yep I missed it :facepalm:. I have rebased this so we can land it.

dark0dave (approved):

stepanstipl:
Looks good, but the commit is not signed -> merging blocked.

dark0dave:
See #519

PR #486, opened 2023-05-10
Scoop install instructions

jmartasek:
https://github.com/ScoopInstaller/Main/pull/4454 has implemented the support for [Scoop](https://scoop.sh). 

Behind the scenes it does the same thing as the install instructions specify (i.e., go to releases and download) plus it adds auto-update too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,12 @@ Or download the
 [latest release](https://github.com/doitintl/kube-no-trouble/releases/latest)
 for your platform and unpack manually.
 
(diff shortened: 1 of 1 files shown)


stepanstipl:
Thanks for letting us know, that's nice. 👍 

We have recently added a 3rd-party install section and note about Homebrew [^1], and would be happy to accept a PR about Scoop in similar fashion.

[^1]: https://github.com/doitintl/kube-no-trouble/pull/501

PR #492, opened 2023-05-31
fix the fixNamespace function

mollonado:
For deprecated resources with no namespace the cmd throws

"

goroutine 1 [running]:
github.com/doitintl/kube-no-trouble/pkg/judge.(*RegoJudge).Eval(0x14000ba8350, {0x1400c258000, 0x394, 0x400})
	/__w/kube-no-trouble/kube-no-trouble/pkg/judge/rego.go:61 +0x684
main.main()
	/__w/kube-no-trouble/kube-no-trouble/cmd/kubent/main.go:146 +0x564
"

That's because the condition in fixNamespaces needs to be reverted to match non-namespaced resources.

### DIFF
--- a/pkg/collector/helm.go
+++ b/pkg/collector/helm.go
@@ -39,7 +39,7 @@ func parseManifests(manifest string, defaultNamespace string, discoveryClient di
 
 func fixNamespace(resource *unstructured.Unstructured, defaultNamespace string, discoveryClient discovery.DiscoveryInterface) {
 	// Default to the release namespace if the manifest doesn't have the namespace set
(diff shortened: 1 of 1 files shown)


stepanstipl:
Sorry for the delay, but I don't think this is desired - we only want to fix namespace for resources that are namespaced, i.e. the condition:
```go
if resource.GetNamespace() == "" && isResourceNamespaced(discoveryClient, resource.GroupVersionKind()) {
```
is good.

We can't fix a namespace for non-namespaced resources, as they do not have a namespace.


The `panic: interface conversion: interface {} is nil, not string` error is a separate thing I would say - or at least the fix should look different if the issue is missing a namespace for non-namespaced resources. I believe it's the same issue as #505.

Please let me know if you think otherwise. Cheers 👍

PR #499, opened 2023-06-26
fix: prioritize KUBECONFIG env over inCluster

MJjainam:
This is the behaviour even for kubectl CLI.

### DIFF
--- a/pkg/collector/kube.go
+++ b/pkg/collector/kube.go
@@ -2,6 +2,7 @@ package collector
 
 import (
 	"fmt"
+	"os"
 
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


chaitanya-rafay (approved):

dark0dave:
this should be handled in config not in the pkg.

stepanstipl:
Hmm.. 🤔 I think it would be best to mirror `kubectl` behaviour in this aspect, as that's what everyone is used to (and I think it's a bit more complex than just checking one env variable). But I would also like to keep it simple, and for sure don't want to reinvent the wheel. Also, there're multiple ways to create the client config in the `client-go` `clientcmd` package.

Can we find out what exactly the `kubectl` does (code ref. would be great)?


[^1]: https://pkg.go.dev/k8s.io/client-go/tools/clientcmd#ClientConfig

MJjainam:
@stepanstipl  I was unable to find the exact way how kubectl loads. Although what I could find was that 
https://github.com/doitintl/kube-no-trouble/blob/1985e721d66e469a591abf122f9bebc04ae7f139/pkg/collector/kube.go#L61 
client-go function does have loading rules from KUBECONFIG env variable.

dark0dave:
this should be handled in config.go

Issue #515, opened 2023-08-14
Multiple additional kinds of the same resource are not respected


atzawada:
When using the additional kinds (`-a`) option, it appears that you cannot check for multiple versions of the same resource in one call to `kubent`. Here is an example, using some cert-manager CRDs:

## Current Behavior

When running `kubent  -a Certificate.v1alpha2.cert-manager.io`:
```
4:22PM INF >>> Kube No Trouble `kubent` <<<
4:22PM INF version 0.7.0 (git sha d1bb4e5fd6550b533b2013671aa8419d923ee042)
4:22PM INF Initializing collectors and retrieving data
4:22PM INF Target K8s version is <Omitted>
4:22PM INF Retrieved <Omitted> resources from collector name=Cluster
4:23PM WRN failed to discover supported resources for networking.k8s.io/v1beta1: the server could not find the requested resource
4:23PM WRN failed to discover supported resources for extensions/v1beta1: the server could not find the requested resource
4:23PM WRN failed to discover supported resources for extensions/v1beta1: the server could not find the requested resource
4:23PM WRN failed to discover supported resources for extensions/v1beta1: the server could not find the requested resource
4:23PM WRN failed to discover supported resources for extensions/v1beta1: the server could not find the requested resource
4:23PM WRN failed to discover supported resources for extensions/v1beta1: the server could not find the requested resource
4:23PM WRN failed to discover supported resources for extensions/v1beta1: the server could not find the requested resource
4:23PM INF Retrieved <Omitted> resources from collector name="Helm v3"
4:23PM INF Loaded ruleset name=custom.rego.tmpl
4:23PM INF Loaded ruleset name=deprecated-1-16.rego
4:23PM INF Loaded ruleset name=deprecated-1-22.rego
4:23PM INF Loaded ruleset name=deprecated-1-25.rego
4:23PM INF Loaded ruleset name=deprecated-1-26.rego
4:23PM INF Loaded ruleset name=deprecated-future.rego
__________________________________________________________________________________________
>>> Additional resources (custom) <<<
------------------------------------------------------------------------------------------
KIND          NAMESPACE        NAME                                  API_VERSION                REPLACE_WITH (SINCE)
Certificate   <Omitted>         <Omitted>                          cert-manager.io/v1alpha2   <na> ()
Certificate   <Omitted>         <Omitted>                          cert-manager.io/v1alpha2   <na> ()
[...]

PR #516, opened 2023-08-14
fix: Properly represent multiple versions of the same custom resource

atzawada:
I took a quick peek into the issue #515 that I opened earlier. Looks like it was an issue with how the Rego template was being rendered for additional custom resources. Simply combined the resource versions in the go code before passing it all off to the template.

If this approach is deemed ok, I will go back and write up some additional testing around this.

### DIFF
--- a/pkg/rules/rego/custom.rego.tmpl
+++ b/pkg/rules/rego/custom.rego.tmpl
@@ -22,9 +22,9 @@ deprecated_resource(r) = api {
 
 deprecated_api(kind, api_version) = api {
 	deprecated_apis = {
-{{- range . }}
-		"{{ .Kind }}": {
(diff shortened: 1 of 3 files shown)


dark0dave:
This approach looks good to me please add tests.

dark0dave:
@atzawada you'll need to rebase this plus sign the commit.

dark0dave (changes requested):
Tests, rebase and signed commit please.

dark0dave (approved):

dark0dave:
Looks like you'll need to format this.

w21froster:
@atzawada this is some functionality I need so I am running your fork -- are you by chance able to get this formatted so it can be merged? 🙏

PR #520, opened 2023-08-17
feat: Add dockerhub to build ci

dark0dave:


stepanstipl:
Can you please elaborate a bit on the reasons behind this one? @dark0dave

In particular, is Dockerhub better than GHCR? I'm not sure we want two maintain 2 registries unless a good reason.

dark0dave:
@stepanstipl I just wanted to make the image more available but you are right.

PR #527, opened 2023-09-05
fix: Return error exit code when collectors fail

### LINKED ISSUE #450: Return code when the cluster is not accessible or if is there an issue with authentication
Hello,

During our tests we found, that _kubent_ returns `return code = 0` both when the cluster does not exist.
```
xzizka@xzizka:~$ kubent -e 
12:39PM INF >>> Kube No Trouble `kubent` <<<
12:39PM INF version 0.7.0 (git sha d1bb4e5fd6550b533b2013671aa8419d923ee042)
12:39PM INF Initializing collectors and retrieving data
12:40PM INF Retrieved 0 resources from collector name=Cluster
12:40PM ERR Failed to retrieve data from collector error="list: failed to list: Get \"https://127.0.0.1:34748/api/v1/secrets?labelSelector=owner%3Dhelm\": dial tcp 127.0.0.1:34748: connect: connection refused" name="Helm v3"
12:40PM INF Loaded ruleset name=custom.rego.tmpl
12:40PM INF Loaded ruleset name=deprecated-1-16.rego
12:40PM INF Loaded ruleset name=deprecated-1-22.rego
12:40PM INF Loaded ruleset name=deprecated-1-25.rego
12:40PM INF Loaded ruleset name=deprecated-1-26.rego
12:40PM INF Loaded ruleset name=deprecated-future.rego
xzizka@xzizka:~$ echo $?
0
```
and when there is an issue with authentication.
```
xzizka@xzizka:~$ kubent -e 
12:40PM INF >>> Kube No Trouble `kubent` <<<
12:40PM INF version 0.7.0 (git sha d1bb4e5fd6550b533b2013671aa8419d923ee042)
12:40PM INF Initializing collectors and retrieving data
12:40PM ERR Failed to initialize collector: <nil> error="failed to assemble client config: error loading config file \"/home/xzizka/.kube/config\": illegal base64 data at input byte 1532"
12:40PM ERR Failed to initialize collector: <nil> error="failed to assemble client config: error loading config file \"/home/xzizka/.kube/config\": illegal base64 data at input byte 1532"
12:40PM INF Loaded ruleset name=custom.rego.tmpl
12:40PM INF Loaded ruleset name=deprecated-1-16.rego
12:40PM INF Loaded ruleset name=deprecated-1-22.rego
12:40PM INF Loaded ruleset name=deprecated-1-25.rego
12:40PM INF Loaded ruleset name=deprecated-1-26.rego
12:40PM INF Loaded ruleset name=deprecated-future.rego
xzizka@xzizka:~$ echo $?
0
```
There is an error message returned about the issue, but I would not expect returning of `return code = 0` when `kubent` was not able to connect successfully to the cluster.
The definition of the `-e` parameter is:
```
-e, --exit-error                      exit with non-zero code when issues are found
```
I would consider the unsuccessful connection to the cluster as an issue.  
Do you agree with this position?

Thank you
Ondrej

dark0dave:
Error, errors out properly closes #450 and increased log level for silent resource failures

### DIFF
--- a/cmd/kubent/main.go
+++ b/cmd/kubent/main.go
@@ -27,27 +27,28 @@ var (
 )
 
 const (
-	EXIT_CODE_SUCCESS      = 0
-	EXIT_CODE_FAIL_GENERIC = 1
-	EXIT_CODE_FOUND_ISSUES = 200
+	EXIT_CODE_SUCCESS             = 0
+	EXIT_CODE_FAIL_GENERIC        = 1
+	EXIT_CODE_FAIL_GET_COLLECTORS = 100
+	EXIT_CODE_FOUND_ISSUES        = 200
 )
 
 func generateUserAgent() string {
 	return fmt.Sprintf("kubent (%s/%s)", version, gitSha)
 }
 
-func getCollectors(collectors []collector.Collector) []map[string]interface{} {
+func getCollectors(collectors []collector.Collector) ([]map[string]interface{}, error) {
 	var inputs []map[string]interface{}
 	for _, c := range collectors {
 		rs, err := c.Get()
 		if err != nil {
 			log.Error().Err(err).Str("name", c.Name()).Msg("Failed to retrieve data from collector")
-		} else {
-			inputs = append(inputs, rs...)
-			log.Info().Str("name", c.Name()).Msgf("Retrieved %d resources from collector", len(rs))
+			return nil, err
 		}
+		inputs = append(inputs, rs...)
+		log.Info().Str("name", c.Name()).Msgf("Retrieved %d resources from collector", len(rs))
 	}
-	return inputs
+	return inputs, nil
 }
 
 func storeCollector(collector collector.Collector, err error, collectors []collector.Collector) []collector.Collector {
@@ -87,7 +88,6 @@ func getServerVersion(cv *judge.Version, collectors []collector.Collector) (*jud
 				if err != nil {
 					return nil, fmt.Errorf("failed to detect k8s version: %w", err)
 				}
-
 				return version, nil
 			}
 		}
@@ -118,12 +118,15 @@ func main() {
 	log.Info().Msg("Initializing collectors and retrieving data")
 	initCollectors := initCollectors(config)
 
-	config.TargetVersion, err = getServerVersion(config.TargetVersion, initCollectors)
+	config.TargetVersion, _ = getServerVersion(config.TargetVersion, initCollectors)
 	if config.TargetVersion != nil {
 		log.Info().Msgf("Target K8s version is %s", config.TargetVersion.String())
 	}
 
-	collectors := getCollectors(initCollectors)
+	collectors, err := getCollectors(initCollectors)
+	if err != nil {
+		os.Exit(EXIT_CODE_FAIL_GET_COLLECTORS)
+	}
 
 	// this could probably use some error checking in future, but
 	// schema.ParseKindArg does not return any error
--- a/cmd/kubent/main_test.go
+++ b/cmd/kubent/main_test.go
@@ -5,7 +5,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -49,7 +48,7 @@ func TestGetCollectors(t *testing.T) {
 	initCollectors := []collector.Collector{}
 	initCollectors = append(initCollectors, fileCollector)
 
-	collectors := getCollectors(initCollectors)
+	collectors, _ := getCollectors(initCollectors)
 
 	if collectors != nil && len(collectors) != 1 {
 		t.Errorf("Did not get file collector correctly with error: %s", err)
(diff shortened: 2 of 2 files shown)


pkg/collector/cluster.go
```diff
@@ -121,7 +121,7 @@ func (c *ClusterCollector) Get() ([]map[string]interface{}, error) {
 		log.Debug().Msgf("Retrieving: %s.%s.%s", g.Resource, g.Version, g.Group)
 		rs, err := ri.List(context.Background(), metav1.ListOptions{})
 		if err != nil {
-			log.Debug().Msgf("Failed to retrieve: %s: %s", g, err)
+			log.Warn().Msgf("Failed to retrieve: %s: %s", g, err)
```
stepanstipl, line 124: I don't think this is desirable, and that we actually previously lowered the log-level of this. It will be a common occurrence that when run on K8S versions that don't already support the resource version, the cluster collector won't be able to retrieve those. And we don't want to warn about that by default.

stepanstipl (commented):
Thanks for this 👍 

The main thing is the desired error behaviour -
I believe that when there's a failure with one of the enabled collectors, we should always exit with an error, irrespective of the `config.ExitError` option. That option/flag is meant to configure whether there should be an error when issues are found - i.e. `kubent` operates successfully, but has found some deprecated resources. If we fail to operate successfully, we should always exit with error.

So in this case, I would error always, irrespective of that flag.



Some minor comments in the code on implementation and the logging bit.

Plus can we cover this with a test? Perhaps it could be another one in https://github.com/doitintl/kube-no-trouble/blob/f6ebd85d8856d41d91578f793587b41a374e1fac/cmd/kubent/main_test.go#L104

cmd/kubent/main.go
```diff
@@ -36,16 +36,19 @@ func generateUserAgent() string {
...
 	var inputs []map[string]interface{}
 	for _, c := range collectors {
 		rs, err := c.Get()
 		if err != nil {
 			log.Error().Err(err).Str("name", c.Name()).Msg("Failed to retrieve data from collector")
-		} else {
-			inputs = append(inputs, rs...)
-			log.Info().Str("name", c.Name()).Msgf("Retrieved %d resources from collector", len(rs))
+			if exitError {
+				os.Exit(EXIT_CODE_FAIL_GENERIC)
```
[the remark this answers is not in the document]
dark0dave: Sure

pkg/collector/cluster.go
dark0dave: ok removed

dark0dave:
@stepanstipl updated please take another look.

dark0dave:
I need a test sec

dark0dave:
@stepanstipl ok, reviewed all comments and updated should be more go like apologies.

cmd/kubent/main_test.go
```diff
@@ -102,7 +101,7 @@ func TestStoreCollectorError(t *testing.T) {
 }
 
 func TestMainExitCodes(t *testing.T) {
-	tmpDir, err := ioutil.TempDir(os.TempDir(), "kubent-tests-")
+	tmpDir, err := os.MkdirTemp(os.TempDir(), "kubent-tests-")
```
dark0dave, line 104: ioutil is deprecated.

cmd/kubent/main.go
```diff
@@ -118,12 +117,15 @@ func main() {
...
-	config.TargetVersion, err = getServerVersion(config.TargetVersion, initCollectors)
+	config.TargetVersion, _ = getServerVersion(config.TargetVersion, initCollectors)
 	if config.TargetVersion != nil {
 		log.Info().Msgf("Target K8s version is %s", config.TargetVersion.String())
 	}
 
-	collectors := getCollectors(initCollectors)
+	collectors, err := getCollectors(initCollectors)
+	if err != nil {
+		os.Exit(EXIT_CODE_FOUND_ISSUES)
```
stepanstipl, line 127:
```suggestion
		os.Exit(EXIT_CODE_FAIL_GENERIC)
```
Should return different - generic fail - exit code.

cmd/kubent/main_test.go
```diff
@@ -128,6 +127,7 @@ func TestMainExitCodes(t *testing.T) {
 		{"version long flag set", []string{"--version"}, 0, "", "", false},
 		{"empty text output", []string{clusterFlagDisabled, helm3FlagDisabled}, 0, "", "", false},
 		{"empty json output", []string{"-o=json", clusterFlagDisabled, helm3FlagDisabled}, 0, "[]\n", "", false},
+		{"fail to initialize collectors", []string{"-o=json", "-f=fail"}, 200, "[]\n", "", false},
```
stepanstipl, line 130: The test is probably OK, but the name is a bit misleading. The error, in this scenario with `-f=fail` flag does not come from collector init, but from the `Get` of the given collector, so maybe just drop the `initialize` bit.

stepanstipl (commented):
I think this is good, just 2 minor nitpicks in code.

But I realised one thing - when looking at the tests. I think the test is good, but I was also thinking it might be good to cover the two scenarios from #450:
a) Nonexisting/wrong cluster endpoint
b) Malformed kubeconfig

And I realised that we probably only cover half of the issue. The 2nd b) failure does not seem to come from `getCollectors`, but rather from the `initCollectors`:
```go
```

Happy to leave the init bit for another PR, and we probablyshould to get this is, but just to be aware we're fixing just half of the scenarios from #450.

cmd/kubent/main.go
dark0dave: Sure will add

stepanstipl (approved):
Hi @dark0dave, you want to get this in?

I'm happy with the code 👍, there's just one test failing (I believe it's just the return code `200 -> 100`). Cheers 🎉

dark0dave:
@stepanstipl should be all good now

Issue #218, opened 2021-09-23
[Feature Request] Running kubent as a webservice


mikhailadvani:
In a multi-tenant kubernetes cluster with tenant not having cluster-admin privileges, sharing the kubent output is not straightforward. Below is my proposal to tackle this:

1. Run kubent as webservice with a relatively privileged serviceAccount - this can be configured by cluster admins
2. Expose the information using a kubernetes service and/or ingress however other tools are exposed
3. Refreshes the deprecation info every n minutes without any restarts
4. Publish helm charts to be able to deploy easily in any cluster

Further extension: expose the information in the form of prometheus metrics so as to be able to create alerts.

Will be more than happy to contribute to this if there is agreement on the idea.

david-doit-intl:
See docker image.

david-doit-intl:
I probably advise running it as a sidecar

david-doit-intl:
We are unlikely to add all of these features as they are a bit beyond what kubent was designed for. Nothing stops you wrapping a service arround our docker image.

mikhailadvani:
Thought about it but personally, I do not like the sidecar pattern primarily because of the orchestration of the periodic refresh and error handling. I am more inclined towards a separate project like a kubent-prometheus-exporter which runs a goroutine periodically to update the list of metrics in memory that are then served by a webserver in prometheus form. A bit of refactoring in the main project here would go a long way in being able to call the entire computation in a single method call without duplication but an os.Exec("kubent...") would suffice as a PoC. I would definitely like to get opinions on the approach

stepanstipl:
Hi @mikhailadvani, thanks for the input. I think turning this into a "proper" webservice, as in real-time listening for API changes and their "stream" processing, and exposing this via web service, is a bit beyond current scope of the tool.

What I would see as an easy implementation of what I think you're trying to achieve is a `CronJob` periodically executing the tool and saving its output somewhere (might be a simple persistent volume). The tool should already run nicely from within the cluster. I think it would make sense to add an example of this, aka deployment manifest to create a relevant service account with limited read-only permissions and the `CronJob` itself.

As to exposing this - the tool already supports JSON, and perhaps an easy way would be to use something like a [json_exporter][1]. It seems flexible enough and might work even without any transformation step. I would suggest grouping findings by resource kind as a label.

[...]

PR #534, opened 2023-09-26
feat: Add prometheus printer

atzawada:
Adding a new "printer" (reporter, exporter?) to push Kubent output to Prometheus via Pushgateway. Any feedback is appreciated.

Relates to #302 and the discussion in #218

### Flags
* `-o prometheus` is now an allowed option

### Environment Variables
* `PUSHGATEWAY_URL`: The URL of the Prometheus Pushgateway.
* `PUSHGATEWAY_JOB`: The job to collect metrics under.

If `-o prometheus` is set, then both of these variables are required to be set.

## Remaining Questions
* Maybe need to expand on the configuration abilities for printers...Is environment variables the right approach or should printers be able to accept custom config/define their own flags?

## Remaining Work
* [ ] Add Pushgateway auth options
* [ ] Add unit testing

### DIFF
--- a/pkg/printer/printer.go
+++ b/pkg/printer/printer.go
@@ -8,9 +8,10 @@ import (
 )
 
 var printers = map[string]func(string) (Printer, error){
-	"json": newJSONPrinter,
-	"text": newTextPrinter,
-	"csv":  newCSVPrinter,
+	"json":       newJSONPrinter,
+	"text":       newTextPrinter,
+	"csv":        newCSVPrinter,
+	"prometheus": newPrometheusPrinter,
 }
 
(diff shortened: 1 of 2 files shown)


isugimpy:
Not to tell you that you shouldn't do this, but you might consider https://github.com/isugimpy/kubent_exporter/ as an alternative to this.  Having a publisher built in would be pretty cool though.

dark0dave:
Hi @atzawada , I think we can support this. Not sure about adding all these env variables. Happy to add the printer, but pushing the messages should probably be handled by another tool.

If you rebase this and add some tests I'll happily merge it.

PR #587, opened 2024-04-16
fix: install.sh in dumb TERM

FabioAntunes:
When $TERM is dumb the number of colours returned is -1 we end up with no values set for the colours.

To test this:
```bash
export TERM=dumb
sh -c "$(curl -sSL https://git.io/install-kubent)"
```

You should get an error regarding parameter expansion:

```
sh: 96: GREEN: parameter not set
```

This PR fixes the issue by initialising the variables no matter what.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -26,6 +26,10 @@ check_binaries() {
 
 # Check if we have colors available, it looks good
 check_colors(){
+  GREEN=''
+  RED=''
+  YELLOW=''
+  NOCOL=''
   if command -v tput > /dev/null; then
     COLORS="$(tput colors)"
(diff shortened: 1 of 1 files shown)


dark0dave:
Ta

dark0dave:
Can you run the pre-commit? @FabioAntunes

FabioAntunes:
I raised the PR through the GitHub ui, noticed that the issue was white
spaces I have pushed some changes to fix those white spaces issue. if
that’s not enough to make the pipeline happy I will run the pre-commit
later today.

dark0dave:
Thanks!

dark0dave (approved):

Issue #505, opened 2023-07-12
panic: interface conversion: interface {} is nil, not string


andyblog:
i get panic when i use tool to check k8s version
k8s version: 1.23.17

```shell
3:40PM INF Loaded ruleset name=deprecated-future.rego
panic: interface conversion: interface {} is nil, not string

goroutine 1 [running]:
github.com/doitintl/kube-no-trouble/pkg/judge.(*RegoJudge).Eval(0xc01ba17920, {0xc002924000, 0x20b9, 0x2400})
        /__w/kube-no-trouble/kube-no-trouble/pkg/judge/rego.go:61 +0x88a
main.main()
        /__w/kube-no-trouble/kube-no-trouble/cmd/kubent/main.go:146 +0x530
``` 

tool version
```shell
./kubent -t 1.24 -l debug -O kubent.log
3:56PM INF >>> Kube No Trouble `kubent` <<<
3:56PM INF version 0.7.0 (git sha d1bb4e5fd6550b533b2013671aa8419d923ee042)
3:56PM INF Initializing collectors and retrieving data
3:56PM INF Target K8s version is 1.24.0
3:56PM DBG Retrieving: daemonsets.v1.apps
```

PR #594, opened 2024-04-24
Warn and Fix Up Invalid Namespaces on Objects

justdan96:
This is to avoid issues such as https://github.com/doitintl/kube-no-trouble/issues/505. I have been seeing this error myself, which manifests as:
```
10:53AM TRC parsing +map[ApiVersion:batch/v1beta1 Kind:CronJob Name:resource123 Namespace:<nil> ReplaceWith:batch/v1 RuleSet:Deprecated APIs removed in 1.25 Since:1.21]
```

This is a simple PR which just warns the user that the object's namespace was not collected correctly and then sets it to `<undefined>`. The message could also be amended to ask the user to report the issue to the developer. I figure this is a better outcome then a panic, and it also fixes the problem I (and presumably others) was having.

### DIFF
--- a/pkg/judge/rego.go
+++ b/pkg/judge/rego.go
@@ -2,6 +2,7 @@ package judge
 
 import (
 	"context"
+
 	"github.com/doitintl/kube-no-trouble/pkg/rules"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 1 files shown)


dark0dave:
You couldn't run the pre-commit and sign this commit please?

dark0dave:
Thanks for the contribution.

justdan96:
> You couldn't run the pre-commit and sign this commit please?

Done and done.

PR #628, opened 2024-08-27
fix(gitcliff): Add fix for git cliff

dark0dave:
Path fix for git cliff

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -155,6 +155,8 @@ jobs:
         with:
           config: cliff.toml
           args: --verbose --latest
+        env:
+          OUTPUT: changelog.md
       - uses: actions/download-artifact@v1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -170,7 +172,7 @@ jobs:
         with:
           tag_name: ${{ steps.get_tag.outputs.git_tag }}
           release_name: ${{ steps.get_tag.outputs.git_tag }}
-          body_path: ./release-artifacts/changelog.md
+          body_path: ./changelog.md
```
apgiorgi, line 175: Would it work with `${{ steps.git-cliff.outputs.changelog }}` ?

dark0dave: Lets find out

apgiorgi (approved):

PR #636, opened 2024-09-25
Fix workflows

apgiorgi:
- Upgrade actions/upload-artifact from v1 to v4
- Upgrade actions/download-artifact from v1 to v4
- Add `migueldelucasdoit` as dependabot reviewer


femrtnz (approved):
lgtm

Issue #410, opened 2022-12-06
FR: Expand deprecation warning to include reason or source


hpl002:
This might be very helm specific, but still.

Had a very old helm revision that triggered a deprecation warning. 
Would be nice if it simply told me the source of this warning as opposed to me having to dig through

Helm encodes the revision directly in the secret name so it should be fairly accessible :D

stepanstipl:
Thanks for the idea 👍 - I was previously against adding too much info in the default output to keep it simple, but at the same time, it might be worth making this more configurable, i.e. custom/extended output format, related to #405.

Issue #405, opened 2022-11-28
Give labels in output properties


mksha:
Hi Guys,

We are using this tool to get deprecated resources but today it only give below fields in output.

Name
Namespace
Kind
ApiVersion
RuleSet
ReplaceWith
Since

I am thinking to add labels as well, so it will be easy to get to know how things are created and can be filtered further based on organization need.

mksha:
@stepanstipl i have tried to add the new fields and it worked but its not returning the all the resources that were returns before adding new additional fields.

can you pls help me to figure out why?

mksha:
@dark0dave @manute @liggitt @meringu @AndreasSko  pls guide me how to contribute

stepanstipl:
Hi @mksha, thanks for the idea 👍 we will for sure consider adding this, it sounds useful.

> @stepanstipl i have tried to add the new fields and it worked but its not returning the all the resources that were returns before adding new additional fields.
>
> can you pls help me to figure out why?

As to this - I or anyone else can't help you without seeing your code 😄 . You can try to share a link to code with changes in your fork & PR against it, but I can't promise to help to guide you on how to implement this feature - it's likely a relatively large change.But hey, worth a try 😄 . Other than that, basic dev requirements are mentioned in a readme [^1], if you have some specific questions to those, let me know.

[^1]: https://github.com/doitintl/kube-no-trouble#development

PR #637, opened 2024-09-27
feat(resource-labels): Add labels flag 

femrtnz:
As part of the issue #405 and and #410 , we are adding a new flag called `--labels`. 
This flag is not active as default 
- to avoid any braking changes. 
- it might be too noise for some people as some resources contain several labels.

Setting the flag `--labels=true` will include in the output a list of labels for each resource, please check the example output below

**Example for TEXT output**

```
>>> Deprecated APIs removed in 1.26 <<<
KIND                      NAMESPACE     NAME            API_VERSION           REPLACE_WITH      (SINCE)     LABELS
HorizontalPodAutoscaler   <undefined>   test-labels-1   autoscaling/v2beta2   autoscaling/v2    (1.23.0)    k8s-app:gcp-compute-persistent-disk-csi-driver
```

**Example for CSV output**

```
api_version,kind,namespace,name,replace_with,since,rule_set,labels
autoscaling/v2beta2,HorizontalPodAutoscaler,<undefined>,test-labels-1,autoscaling/v2,1.23.0,Deprecated APIs removed in 1.26,k8s-app:gcp-compute-persistent-disk-csi-driver
```

**Example for JSON output**

```json
[
	{
		"Name": "test-labels-1",
		"Namespace": "\u003cundefined\u003e",
		"Kind": "HorizontalPodAutoscaler",
		"ApiVersion": "autoscaling/v2beta2",
		"RuleSet": "Deprecated APIs removed in 1.26",
		"ReplaceWith": "autoscaling/v2",
		"Since": "1.23.0",
		"Labels": {
			"k8s-app": "gcp-compute-persistent-disk-csi-driver"
		}
	}
]
```

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1,9 +1,11 @@
 package config
 
 import (
-	goversion "github.com/hashicorp/go-version"
+	"context"
 	"os"
 	"testing"
+
+	goversion "github.com/hashicorp/go-version"
 
 	"github.com/spf13/pflag"
 )
@@ -11,6 +13,7 @@ import (
 func TestValidLogLevelFromFlags(t *testing.T) {
 	oldArgs := os.Args[1]
 	defer func() { os.Args[1] = oldArgs }()
+	ctx := context.Background()
 
 	var validLevels = []string{"trace", "debug", "info", "warn", "error", "fatal", "panic", "", "disabled"}
 	for i, level := range validLevels {
@@ -18,7 +21,8 @@ func TestValidLogLevelFromFlags(t *testing.T) {
 		pflag.CommandLine = pflag.NewFlagSet(os.Args[0], pflag.ExitOnError)
 
 		os.Args[1] = "--log-level=" + level
-		config, err := NewFromFlags()
+
+		config, _, err := NewFromFlags(ctx)
 
 		if err != nil {
 			t.Errorf("Flags parsing failed %s", err)
@@ -44,8 +48,9 @@ func TestInvalidLogLevelFromFlags(t *testing.T) {
 func TestNewFromFlags(t *testing.T) {
 	// reset for testing
 	pflag.CommandLine = pflag.NewFlagSet(os.Args[0], pflag.ExitOnError)
+	ctx := context.Background()
 
-	config, err := NewFromFlags()
+	config, _, err := NewFromFlags(ctx)
 
 	if err != nil {
 		t.Errorf("Flags parsing failed %s", err)
@@ -72,9 +77,9 @@ func TestValidateAdditionalResources(t *testing.T) {
 
 func TestValidateAdditionalResourcesFail(t *testing.T) {
 	testCases := [][]string{
-		[]string{"abcdef"},
-		[]string{""},
-		[]string{"test.v1.com"},
+		{"abcdef"},
+		{""},
+		{"test.v1.com"},
 	}
 
 	for _, tc := range testCases {
@@ -90,6 +95,7 @@ func TestTargetVersion(t *testing.T) {
 	validVersions := []string{
 		"1.16", "1.16.3", "1.2.3",
 	}
+	ctx := context.Background()
 
 	oldArgs := os.Args[1]
 	defer func() { os.Args[1] = oldArgs }()
@@ -99,7 +105,7 @@ func TestTargetVersion(t *testing.T) {
 		pflag.CommandLine = pflag.NewFlagSet(os.Args[0], pflag.ExitOnError)
 
 		os.Args[1] = "--target-version=" + v
-		config, err := NewFromFlags()
+		config, _, err := NewFromFlags(ctx)
 
 		if err != nil {
 			t.Errorf("Flags parsing failed %s", err)
@@ -120,7 +126,7 @@ func TestTargetVersionInvalid(t *testing.T) {
 	invalidVersions := []string{
 		"1.blah", "nope",
 	}
-
+	ctx := context.Background()
 	oldArgs := os.Args[1]
 	defer func() { os.Args[1] = oldArgs }()
 
@@ -129,7 +135,7 @@ func TestTargetVersionInvalid(t *testing.T) {
 		pflag.CommandLine = pflag.NewFlagSet(os.Args[0], pflag.ContinueOnError)
 
 		os.Args[1] = "--target-version=" + v
-		config, _ := NewFromFlags()
+		config, _, _ := NewFromFlags(ctx)
 
 		if config.TargetVersion != nil {
 			t.Errorf("expected --target-version flag parsing to fail for: %s", v)
@@ -141,7 +147,7 @@ func TestContext(t *testing.T) {
 	validContexts := []string{
 		"my-context",
 	}
-
+	ctx := context.Background()
 	oldArgs := os.Args[1]
 	defer func() { os.Args[1] = oldArgs }()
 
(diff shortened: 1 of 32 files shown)


femrtnz:
Hey @dark0dave I appreciate any insights on this change 🙏

pkg/config/config_test.go
```diff
@@ -72,9 +72,9 @@ func TestValidateAdditionalResources(t *testing.T) {
 
 func TestValidateAdditionalResourcesFail(t *testing.T) {
 	testCases := [][]string{
-		[]string{"abcdef"},
-		[]string{""},
-		[]string{"test.v1.com"},
```
eyalzek, line 77: Why remove types here?

pkg/printer/csv.go
eyalzek, line 1: Using a variable for printed header/fields might be cleaner (i.e. append a value in case of the labels switch)

pkg/printer/printer.go
```diff
@@ -14,7 +14,7 @@ var printers = map[string]func(string) (Printer, error){
 }
 
 type Printer interface {
-	Print([]judge.Result) error
+	Print([]judge.Result, bool) error
```
eyalzek, line 17: I'm wondering if it would make sense to pass an object than contains optional flags, rather than just on unclear `bool` here

eyalzek (commented):
Left some comments as I'm not familiar enough with the project nor the style guide :)

pkg/config/config_test.go
femrtnz: This is a redundant type from array composite literal, so doesn't need to be there.

pkg/printer/csv.go
femrtnz: Yeah agree! Fixed at https://github.com/doitintl/kube-no-trouble/pull/637/commits/f5c9ea6d61ac6f1674403e23c10276530c880ac5

pkg/printer/printer.go
femrtnz: Yeah agree. Let' me check the effort on changing this.

migueldelucasdoit (approved):
Hey @femrtnz can we have a look at the duplicated code we introduced in this PR?
https://github.com/doitintl/kube-no-trouble/pull/637/checks?check_run_id=30865045590 

We can refactor the code to extract an internal function with the duplicated code.
https://www.jetbrains.com/help/go/extract-method.html

pkg/printer/printer.go
femrtnz: done at https://github.com/doitintl/kube-no-trouble/pull/637/commits/e7e6812e916c784c3283407dab3690b7d2c3d129

dark0dave:
This PR is too long, style is incorrect and introduces way too many commits. Rebase and clean this up.

dark0dave (commented):
Clean this up. Or split into seperate PRS.

dark0dave:
Judge should *not* depend on config. Please don't create this binding....

femrtnz:
> Judge should _not_ depend on config. Please don't create this binding....

Could you please detail what you mean? which file, line etc? it will be easier if you comment on the file.

dark0dave:
Also why has the printer changed.... Like this feels brittle.

pkg/printer/csv.go
dark0dave, line 1: Here printer should not know about config.

femrtnz: Any design suggestion to manipulate the outputs based on flags?

We could create a middle layer DTO  e convert the file over but it might be a new unnecessary new layer + complexity.

Let me know your thoughts.

femrtnz: alternatively I can add basic types to the `outputResults` function, ie `bool` however we might need to add more parameters and might not be the best solution

femrtnz: Hey @dark0dave any thoughts on this ^^

femrtnz: FYI Instead of adding config to printer I'm including this to the context

pkg/context/context-keys.go
```diff
@@ -0,0 +1,5 @@
+package context
+
+type ctxKey string
+
+const LEBELS_CTX_KEY ctxKey = "labels"
```
migueldelucasdoit, line 5: Please @femrtnz fix this typo `LABELS_CTX_KEY`

pkg/printer/csv_test.go
```diff
@@ -50,10 +51,12 @@ func TestCSVPrinterPrint(t *testing.T) {
 		commonPrinter: &commonPrinter{tmpFile},
 	}
 
-	version, _ := judge.NewVersion("1.2.3")
-	results := []judge.Result{{"Name", "Namespace", "Kind", "1.2.3", "Test", "4.5.6", version}}
+	labelsFlag := false
+	ctx := context.WithValue(context.Background(), ctxKey.LEBELS_CTX_KEY, &labelsFlag)
```
migueldelucasdoit, line 55: Please @femrtnz fix this typo `LABELS_CTX_KEY`

pkg/printer/printer_helper_test.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+			defer os.Remove(tmpFile.Name())
+
+			tp := &csvPrinter{
+				commonPrinter: &commonPrinter{tmpFile},
+			}
+
+			results := getTestResult(tt.labels)
+
+			ctx := context.WithValue(context.Background(), ctxKey.LEBELS_CTX_KEY, &tt.withLabels)
```
migueldelucasdoit, line 44: Please @femrtnz fix this typo `LABELS_CTX_KEY`

pkg/printer/text.go
```diff
@@ -48,15 +50,26 @@ func (c *textPrinter) Print(results []judge.Result) error {
 	ruleSet := ""
 	w := tabwriter.NewWriter(c.commonPrinter.outputFile, 10, 0, 3, ' ', 0)
 
+	labels := ctx.Value(ctxKey.LEBELS_CTX_KEY).(*bool)
```
migueldelucasdoit, line 53: I think it would be a good idea to use a helper function(s) to manage the data in the context WDYT @femrtnz 
```
func GetLabels(ctx context.Context) (*bool, error) {
	if v := ctx.Value(LABELS_CTX_KEY); v != nil {
		return ctx.Value(LABELS_CTX_KEY).(*bool), nil
	}
	return nil, error.New("labels not found in context.")
}
```

migueldelucasdoit (approved):
Just some typos to fix and some recommendations.

pkg/printer/csv.go
```diff
@@ -46,26 +48,40 @@ func (c *csvPrinter) Print(results []judge.Result) error {
...
 		"kind",
 		"namespace",
 		"name",
 		"replace_with",
 		"since",
 		"rule_set",
-	})
+	}
+
+	labels := ctx.Value(ctxKey.LEBELS_CTX_KEY).(*bool)
```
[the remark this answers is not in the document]
femrtnz: thanks!

pkg/printer/text.go
femrtnz: yeah agree! fizex on last commit

dark0dave:
> **Printer** should _not_ depend on config. Please don't create this binding....

@femrtnz 
Read this again and now understand my pr https://github.com/doitintl/kube-no-trouble/pull/649

dark0dave:
This pr is massive mistake. Not the functionality the way it was done.

PR #644, opened 2024-10-17
Really clean during make all

LeMyst:


femrtnz (approved):
Thanks for contributing!

PR #649, opened 2024-10-25
fix(context): Removing unneeded global context

dark0dave:
Please do not merge this until its ready.


migueldelucasdoit:
Hello @dark0dave we would like to find the best way to collaborate together. Can you please explain in more detail what you're trying to do with this PR? Happy to discuss this with you in person and with the rest of the team.

dark0dave:
> Hello @dark0dave we would like to find the best way to collaborate together. Can you please explain in more detail what you're trying to do with this PR? Happy to discuss this with you in person and with the rest of the team.

I mean now your happy to collborate???

dark0dave:
Standards in this project have been either ignored or just gone unnoticed.

dark0dave:
I'll state it for brevity again. Do not introduce unneed global dependancies...

dark0dave:
This still isn't ready please don't  merge, I need to correct the tests.

femrtnz:
@dark0dave please let us know when this is ready to review, thanks!

dark0dave:
@femrtnz @migueldelucasdoit this pr is now ready. In future if we could avoid introducing global context varriables without discuss that would be great. So I don't have to spend my free time correcting issues.

image

dark0dave:
Maintianing this repo is not a race.

dark0dave:
The repeated code is required for tests. We can clean that up later. Can I get a review from @migueldelucasdoit @femrtnz.

The intent here is as I have said in line with removing global context. In an ideal situtation we do not share a global context. So instead I have introduced cusomizable printer options as I am sure many more printing requests will arise. 

This leaves room for the future, rather than placing in a global context, which created a tight binding accross the sub packages. I know this is the natural inclination of golang to add a context, but we want to keep these packages seperate. Your welcome to introduce a context for indivual packages, see pkg/judge/rego.go for a nice example of how this is done. 

We try to use the config package to hold configuration for the whole of kubent, so by introducing a context you effectively stored configuration in two places which is not desirable. Please try to keep this in mind when adding changes.

femrtnz (approved):
Hey @dark0dave thanks for detailed explanation. I agree it is a better approach removing the global context.

Whats is your plan to fix the tests?  Looking the duplications I can see benefits on having those duplicated there to improve readability. Maybe we could ignore duplications on test files in Sonar? Or create a some helper functions?

dark0dave:
> Hey @dark0dave thanks for detailed explanation. I agree it is a better approach removing the global context.
> 
> Whats is your plan to fix the tests? Looking the duplications I can see benefits on having those duplicated there to improve readability. Maybe we could ignore duplications on test files in Sonar? Or create a some helper functions?

Lets solve the duplications in annother pr.  I have created an issue to track it. Open to suggestions!

dark0dave:
https://github.com/doitintl/kube-no-trouble/issues/652

PR #659, opened 2024-11-15
ref: Add OutputFormat type to enforce flag validation at parse time

harry97uk:
I have added an OutputFormat type in a separate file within the config pkg.  I have also added a couple of new unit tests to validate that the type is working as intended.

I set the default for the printer type as TEXT and removed code that did the printer type validation in config.go

I did not change anything in the printer pkg as I wanted to keep this PR small.


dark0dave:
Looks like the pre-commit failed you may have to run:

```sh
go fmt . /...
```

There also seems some slight spelling mistakes.

If you can clean these up will get this merged.

Thank you again.

